PR #6, opened 2016-05-01
Async Support For Prefork

FreddieWitherden:
This request cleans up the preforking module and adds support for asynchronous invocations.  This is accomplished by three new methods: `call_async`, `wait` and `waitall`.  The deprecated method `call_capture_stdout` has also been removed.

Note that it is not currently possible to capture the output from asynchronous processes.  Doing this right is somewhat fiddly as the OS will only buffer so much output from a process before blocking it (waiting on the parent to `.communicate` with it).  This is difficult to accomplish without moving towards threading which would be an additional complication.

Note also that the module is still thread safe.  Only one thread from the parent may call into the prefork routines at any one time.  While this can be worked around it probably does not warrant the additional complexity.

### DIFF
--- a/pytools/prefork.py
+++ b/pytools/prefork.py
@@ -13,18 +13,36 @@ class ExecError(OSError):
 
 
 class DirectForker(object):
-    @staticmethod
-    def call(cmdline, cwd=None):
+    def __init__(self):
+        self.apids = {}
+        self.count = 0
+
+    def call(self, cmdline, cwd=None):
         from subprocess import call
+
         try:
             return call(cmdline, cwd=cwd)
         except OSError as e:
             raise ExecError("error invoking '%s': %s"
                             % (" ".join(cmdline), e))
 
-    @staticmethod
-    def call_capture_output(cmdline, cwd=None, error_on_nonzero=True):
+    def call_async(self, cmdline, cwd=None):
+        from subprocess import Popen
+
+        try:
+            self.count += 1
+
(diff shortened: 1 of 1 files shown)


inducer:
Looks good. Thanks for the contribution, and sorry that about not having had a chance to respond to your original email.

FreddieWitherden:
Great, thanks!

Would you be able to mint a new release so this makes it up onto PyPI?

inducer:
Done.

inducer:
Uh-oh. This needs to be fixed. The compatibility target for pytools is still Python 2.6. Could you send a PR to mop this up?

```
Traceback (most recent call last):
  File "/home/gitlab-runner/builds/adb2419f/0/inducer/meshpy/test/test_meshpy.py", line 215, in test_gmsh
    generate_gmsh(mr, GMSH_SPHERE, 3)
  File "build/bdist.linux-x86_64/egg/meshpy/gmsh_reader.py", line 439, in generate_gmsh
    runner.__enter__()
  File "build/bdist.linux-x86_64/egg/meshpy/gmsh.py", line 115, in __enter__
    from pytools.prefork import call_capture_output
  File "/home/gitlab-runner/builds/adb2419f/0/inducer/meshpy/.env/lib/python2.6/site-packages/pytools-2016.2-py2.6.egg/pytools/prefork.py", line 68
    return {aid: self.wait(aid) for aid in list(self.apids)}
                                  ^
SyntaxError: invalid syntax
```

PR #7, opened 2016-05-17
Fix prefork for Python 2.6.

FreddieWitherden:


inducer:
Thanks for getting this done so quickly! I'll do a point relelase.

inducer:
Done.

PR #9, opened 2016-11-29
[bugfix] Use base classes for ImmutableRecord in correct order

dokempf:
A very weird bug: The order of base classes actually matters here,
because if __hash__ and __eq__ are defined in the wrong order,
__hash__ will get deleted (in 3). MWE:

```
class X(object):
    def __eq__(self, other):
        return True

class Y(object):
    def __hash__(self):
        return 0

class Z1(Y,X):
    pass

class Z2(X,Y):
    pass

hash(Z1())
hash(Z2())
```

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -207,7 +207,7 @@ class ImmutableRecordWithoutPickling(RecordWithoutPickling):
                     for field in self.__class__.fields))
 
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for figuring this out! I'll release a new version.

inducer:
(Done.)

Issue #15, opened 2019-10-23
'NoneType' object has no attribute 'isatty'


xywei:
When I run commands in the interactive shell of `pudb`, `sys.stdin` is `None`. Thus the following line in `__init__.py` makes it impossible to call certain commands within `pudb` (e.g., `wrangler.translate_box_multipoles_to_qbx_local(mpole_exps)) ` from `pytential`).
```
use_late_start_logging = sys.stdin.isatty()   
```

I am happy to submit a patch if this is intended to be fixed.

inducer:
Yup, please do submit a patch!

xywei:
I am closing this since the original issue has been fixed.

PR #16, opened 2019-10-23
Set use_late_start_logging=False if stdin is None

xywei:
This should fix the issue described in #15.

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -2182,7 +2182,11 @@ class ProcessLogger(object):  # pylint: disable=too-many-instance-attributes
         # feature for interactive use, do not do it unless there is (weak)
         # evidence of interactive use.
         import sys
-        use_late_start_logging = bool(sys.stdin and sys.stdin.isatty())
+        if sys.stdin is None:
+            # Can happen, e.g., if pudb is controlling the console.
+            use_late_start_logging = False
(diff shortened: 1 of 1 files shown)


pytools/__init__.py
```diff
@@ -2182,7 +2182,7 @@ def __init__(  # pylint: disable=too-many-arguments
         # feature for interactive use, do not do it unless there is (weak)
         # evidence of interactive use.
         import sys
-        use_late_start_logging = sys.stdin.isatty()
+        use_late_start_logging = bool(sys.stdin and sys.stdin.isatty())
```
inducer: This is super hard to read IMO---or easy to read incorrectly. Just spell it out:
```py
if sys.stdin is None:
    # Can happen, e.g., if pudb is controlling the console.
    use_late_start_logging = False
else sys.stdin.isatty():
    use_late_start_logging = sys.stdin.isatty()
```

xywei: OK, I have replaced the line.

xywei:
The CI failures seem to be caused by networking errors when trying to downloading the script.

inducer:
Whoops. Looks like that ran *just* as I was updating gitlab.

inducer:
Rerunning now.

inducer:
Thanks!

PR #21, opened 2020-06-03
Mypy support + PEP 561 support

mattwala:
Following this: https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages
This is not currently complete. I was imagining we could add more stubs as needed. Another option would be to use something like [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html), but its output could be improved (it won't add generic types, for instance).


mattwala:
No idea how to test this :/

inducer:
> No idea how to test this :/

Mypy?

inducer:
Let's drop Py2 and stick these annotations straight into the source.

mattwala:
> > No idea how to test this :/
> 
> Mypy?

I don't actually know whether Mypy can validate stub files against the actual code. If we switch to inline annotations, then we could use Mypy.

mattwala:
In dfac040f4ea992ea03ba7c616e51c8ccb3eaa166 I've dropped Py2 support and added inline type annotations and a Mypy CI job.

Corresponding Gitlab CI results: https://gitlab.tiker.net/inducer/pytools/-/merge_requests/38

mattwala:
There were a few files that had Mypy errors. I've ignored them for now in `setup.cfg`, except for the errors in `__init__.py` which were fixed.

inducer:
LGTM. Thanks for working on this!

PR #24, opened 2020-06-12
setup.py: Drop 3.4, 3.5 support

mattwala:
`__init__.py` uses PEP 526 variable annotations, which are supported only in 3.6+


inducer:
Thx!

PR #29, opened 2020-07-20
Topological ordering with priorities

kaushikcfd:


pytools/graph.py
```diff
@@ -178,24 +178,31 @@ def __init__(self, node):
...
-def compute_topological_order(graph):
+def compute_topological_order(graph, key=None):
     """Compute a toplogical order of nodes in a directed graph.
 
     :arg graph: A :class:`collections.abc.Mapping` representing a directed
         graph. The dictionary contains one key representing each node in the
         graph, and this key maps to a :class:`collections.abc.Iterable` of
         nodes that are connected to the node by outgoing edges.
 
+    :arg key: A custom key function may be supplied to determine the order in
```
inducer, line 189: Should document the calling signature of the function.

pytools/graph.py
```diff
@@ -232,7 +239,38 @@ def compute_topological_order(graph):
                 # so either this is a leaf or all children have been visited
                 visiting.remove(node)
                 visited.add(node)
-                reverse_order.append(node)
+
+                if key is None:
+                    reverse_order.append(node)
+                else:
+                    # {{{ move every 'u' from heap to reverse_order s.t. 'node'->'u'
```
inducer, line 246: What is `u`? What does the `->` arrow mean?

pytools/graph.py
```diff
@@ -178,24 +178,31 @@ def __init__(self, node):
...
     """Compute a toplogical order of nodes in a directed graph.
 
     :arg graph: A :class:`collections.abc.Mapping` representing a directed
         graph. The dictionary contains one key representing each node in the
         graph, and this key maps to a :class:`collections.abc.Iterable` of
         nodes that are connected to the node by outgoing edges.
 
+    :arg key: A custom key function may be supplied to determine the order in
+        break-even cases. If provided the nodes of the graph must be comparable
+        and hashable.
```
inducer, line 191:
```suggestion
        break-even cases. If provided the nodes of the graph must be hashable
        (and hence equality-comparable).
```
("comparable" is different from "equality-comparable")

pytools/graph.py
```diff
@@ -178,24 +178,31 @@ def __init__(self, node):
...
     .. versionadded:: 2020.2
     """
+    from heapq import heappop, heappush
 
     # find a valid ordering of graph nodes
     reverse_order = []
     visited = set()
     visiting = set()
+    heap = []
+    nodes_in_heap = set()
```
inducer, line 205:
```suggestion
    if key is not None:
        heap = []
        nodes_in_heap = set()
```

test/test_graph_tools.py
```diff
@@ -222,6 +222,24 @@ def test_induced_subgraph():
...
+            'c': ['d', 'e'],
+            'd': [],
+            'e': []}
+
+    def key(u):
+        return priorities[u]
+
+    assert compute_topological_order(dag, key=key) == ['a', 'c', 'e', 'b', 'd']
+
+
```
inducer, line 242: I don't find this test convincing. Sure, it can live. But I would prefer a randomly generated graph with randomly generated priorities and a test that verifies the invariant satisfied by the topsort.

pytools/graph.py
```diff
@@ -232,7 +239,38 @@ def compute_topological_order(graph):
...
+                    while nodes_to_move:
+                        node_to_move = nodes_to_move.pop()
+
+                        while True:
+                            _, popped_node = heappop(heap)
+                            reverse_order.append(popped_node)
+                            nodes_to_move.discard(popped_node)
+                            nodes_in_heap.remove(popped_node)
+                            if popped_node == node_to_move:
+                                break
```
inducer, line 259: What is the situation here?

pytools/graph.py
```diff
@@ -178,24 +178,31 @@ def __init__(self, node):
...
 
     .. versionadded:: 2020.2
     """
+    from heapq import heappop, heappush
 
     # find a valid ordering of graph nodes
     reverse_order = []
     visited = set()
     visiting = set()
+    heap = []
```
inducer, line 204: Explain the role of the heap.

pytools/graph.py
kaushikcfd: It was supposed to mean there exists an edge from `node` to `u` in the `graph`. Looks like it needs some clarification.

pytools/graph.py
kaushikcfd: Since `node` is ready to be scheduled i.e. its time to push `node` into `heap`. `heap` cannot contain any immediate successors of `node` as its only job is to help maintaining  priorities but not in a way that invalidates the topological sort order. So all the immediate successors of `node` that are present the `heap` must be completely scheduled (i.e. moved to reverse_order).

The job of the while-loop+heap is to completely schedule all the nodes that are at lesser priority than the successors of `node` in that order.

inducer: (Don't explain these things in PR comments. Add actual code comments.)

pytools/graph.py
kaushikcfd: Resolving as that part of code no longer exists.

pytools/graph.py
kaushikcfd: Resolving as that part of code no longer exists.

test/test_graph_tools.py
kaushikcfd: Added some more generalized tests.

pytools/graph.py
kaushikcfd: Resolving as that part of code no longer exists.

pytools/graph.py
kaushikcfd: Resolving as that part of code no longer exists.

pytools/graph.py
kaushikcfd: Done.

pytools/graph.py
```diff
@@ -236,6 +237,74 @@ def compute_topological_order(graph):
...
+
+    :arg key: A custom key function may be supplied to determine the order in
+        break-even cases.
+
+    :returns: A :class:`list` representing a valid topological ordering of the
+        nodes in the directed graph.
+
+    .. note::
+
+        * Has stricter requirements on the provided *graph* than
```
kaushikcfd, line 258: Was forced to introduce `compute_topological_order_v2` because of this added requirement.

The hashability requirement of graph's keys sprung up because we are now implementing Kahn's algorithm, which was in-turn required to implement the prioritized topological sort order. The earlier implementation was broken and I was left with the conclusion that extending the DFS of `compute_topological_order` to prioritized topological sort cannot be done without hurting the time complexity.

kaushikcfd:
Ready for another round of review.

pytools/graph.py
```diff
@@ -236,6 +237,75 @@ def compute_topological_order(graph):
 
     return list(reversed(reverse_order))
 
+
+def compute_topological_order_v2(graph, key=None):
+    """Compute a toplogical order of nodes in a directed graph.
```
inducer, line 242: Could you link to a source for Kahn's algorithm/state in the code that that's what's being used?

pytools/graph.py
```diff
@@ -35,6 +35,7 @@
 .. autofunction:: compute_sccs
 .. autoclass:: CycleError
 .. autofunction:: compute_topological_order
```
inducer, line 37: I think we should deprecate `compute_topological_order` then.

pytools/graph.py
inducer: As discussed during meeting today, I don't think what's required of the graph keys is actually stronger than before, which means that we can have this replace `compute_topological_order`.

pytools/graph.py
```diff
@@ -236,6 +237,75 @@ def compute_topological_order(graph):
...
+        extract a comparison key from each node of the *graph*.
+
+    :returns: A :class:`list` representing a valid topological ordering of the
+        nodes in the directed graph.
+
+    .. note::
+
+        * Has stricter requirements on the provided *graph* than
+          :func:`pytools.graph.compute_topological_order`: it requires the
+          keys of the mapping *graph* to be hashable and inequality comparable.
```
inducer, line 260: What's "inequality comparable"?

pytools/graph.py
```diff
@@ -236,6 +237,75 @@ def compute_topological_order(graph):
...
+        * Implements Kahn's algorithm.
```
inducer, line 261:
```suggestion
        * Implements `Kahn's algorithm 
          <https://en.wikipedia.org/w/index.php?title=Topological_sorting&oldid=957499939#Kahn's_algorithm>`__.
```

pytools/graph.py
```diff
@@ -236,6 +237,75 @@ def compute_topological_order(graph):
...
+            nodes_to_num_predecessors[child] += 1
+
+    # }}}
+
+    # heap: list of pairs of the form (key(n), n) where 'n' is a node in
+    # 'graph' with no predecessor. Nodes with no predecessors are the
+    # schedulable candidates.
+    heap = [(key(n), n) for n, num_preds in nodes_to_num_predecessors.items()
+            if num_preds == 0]
+    heapify(heap)
```
inducer, line 287: I suspect this might fail if `n1 < n2` is not defined/errors for two nodes. (Since tuples compare lexicographically.) This is easily fixed with a custom class that overrides comparison.

test/test_graph_tools.py
```diff
@@ -222,6 +222,78 @@ def test_induced_subgraph():
...
+            # to avoid cycles only consider edges node_i->node_j where j > i.
+            for j in range(i+1, nnodes):
+                # Edge probability 4/n: Generates decently interesting inputs.
+                if rng.randint(0, nnodes - 1) <= 2:
+                    graph[i].add(j)
+        return graph
+
+    nnodes = rng.randint(40, 100)
+    rev_dep_graph = generate_random_graph(nnodes)
+    dep_graph = dict((i, set()) for i in range(nnodes))
```
inducer, line 269:
```suggestion
    dep_graph = {i: set() for i in range(nnodes)}
```

pytools/graph.py
```diff
@@ -178,63 +179,85 @@ def __init__(self, node):
...
 
-def compute_topological_order(graph):
-    """Compute a toplogical order of nodes in a directed graph.
+class HeapEntry:
+    def __init__(self, node, key):
+        self.node = node
+        self.key = key
+
+    def __lt__(self, other):
+        return self.key < other.key
```
inducer, line 188: Could you add a comment pointing to docs that that's all that's needed for `heapq`? (Aka why doesn't this need the full assortment of `__eq__`, `__le__`, etc?)

test/test_graph_tools.py
inducer: Thanks! Those look good.

inducer:
One last concern, then ready to go.

pytools/graph.py
kaushikcfd: Didn't find any mentions for it in the docs. However the code for `heapq` says it. See https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Lib/heapq.py#L135-L138

kaushikcfd: Didn't find any mentions for it in the docs. However the code for `heapq` says it. See https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Lib/heapq.py#L135-L138

inducer:
LGTM. Thanks for working on this!

PR #30, opened 2020-07-28
Add pytools.codegen

mattwala:
This change takes common code generation functionality out of
pytools.py_codegen and makes it a separate module.

A language-independent parent class of PythonCodeGenerator is also
added.

The version is bumped to 2020.4

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -9,6 +9,7 @@ Welcome to pytools's documentation!
     obj_array
     persistent_dict
     graph
+    codegen
     misc
 
 Indices and tables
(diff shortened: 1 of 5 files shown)


mattwala:
Context: Thought this would be useful in pytato for the graph visualization.

mattwala:
Would also be useful in dagrt, because FortranEmitter currently inherits from PythonCodeGenerator :)

inducer:
Sure, sounds good. Thanks for working on this!

mattwala:
Thanks! Forgot to add a file: #31

PR #32, opened 2020-08-25
add github-flavored markdown and aligned output for Table

matthiasdiener:
See https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables
for specification.

- Outputs a table that is a valid GFM table, and gets rendered automatically as a table on GitHub.
- Adds column alignment specification and makes output respect it.

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1513,6 +1513,21 @@ class Table:
         return "\n".join(lines)
 
     def github_markdown(self):
+        """Returns a string representation of the table formatted as
+        GitHub-Flavored Markdown.
+        https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables
+
+        .. doctest ::
+
+            >>> tbl = Table(alignments=['l', 'r'])
+            >>> tbl.add_row([1, 2])
+            >>> tbl.add_row([10, 20])
+            >>> s = tbl.github_markdown().splitlines()
+            >>> assert s[0] == "1  |  2"
+            >>> assert s[1] == ":--|---:"
+            >>> assert s[2] == "10 | 20"
+
+        """
         columns = len(self.rows[0])
         col_widths = [max(len(row[i]) for row in self.rows)
                       for i in range(columns)]
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1527,16 +1527,19 @@ class Table:
         .. doctest ::
 
             >>> tbl = Table(alignments=['l', 'r'])
-            >>> tbl.add_row([1, 2])
+            >>> tbl.add_row([1, '|'])
             >>> tbl.add_row([10, 20])
             >>> s = tbl.github_markdown().splitlines()
-            >>> assert s[0] == "1  |  2"
+            >>> assert s[0] == "1  | \|"
             >>> assert s[1] == ":--|---:"
             >>> assert s[2] == "10 | 20"
 
         """
-        columns = len(self.rows[0])
-        col_widths = [max(len(row[i]) for row in self.rows)
+        # Pipe symbols ('|') must be replaced
+        rows = [[w.replace('|', '\\|') for w in r] for r in self.rows]
+
+        columns = len(rows[0])
+        col_widths = [max(len(row[i]) for row in rows)
                       for i in range(columns)]
 
         alignments = self.alignments
@@ -1549,7 +1552,7 @@ class Table:
             else cell.ljust(col_width) if align == "l"
             else cell.rjust(col_width)
             for cell, col_width, align in zip(row, col_widths, alignments)])
-            for row in self.rows]
+            for row in rows]
         lines[1:1] = ["|".join(
             ":" + "-" * (col_width - 1 + (i > 0)) + ":" if align == "c"
             else ":" + "-" * (col_width + (i > 0)) if align == "l"
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1546,8 +1546,8 @@ class Table:
 
     def github_markdown(self):
         """Returns a string representation of the table formatted as
-        GitHub-Flavored Markdown.
-        https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables
+        `GitHub-Flavored Markdown.
+        <https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables>`__
 
         .. doctest ::
 
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1499,9 +1499,12 @@ class Table:
     .. automethod:: github_markdown
     """
 
-    def __init__(self, alignments=['l', 'r']):
+    def __init__(self, alignments=None):
         self.rows = []
-        self.alignments = alignments
+        if alignments is not None:
+            self.alignments = alignments
+        else:
+            self.alignments = ['l']
 
     def add_row(self, row):
         self.rows.append([str(i) for i in row])
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1487,11 +1487,10 @@ class Table:
     """An ASCII table generator.
 
     :arg alignments: List of alignments of each column ('l', 'c', or 'r',
-        for left, center, and right alignment, respectively). Currently only used
-        by the `github_markdown` output formatter. Columns which have no alignment
-        specifier will use the last specified alignment. For example, with
-        `alignments=['l', 'r']`, the third and all following columns will use 'r'
-        alignment.
+        for left, center, and right alignment, respectively). Columns which
+        have no alignment specifier will use the last specified alignment. For
+        example, with `alignments=['l', 'r']`, the third and all following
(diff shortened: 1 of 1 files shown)


matthiasdiener:
here is how it looks like:

```python
import pytools

x=pytools.Table(alignments=['l', 'c', 'r'])

x.add_row([1000,2,3,4,5,6,7])
x.add_row([1,2,3,4,5,6,7])
x.add_row([1,2000,3,4,5,6,7])
x.add_row([1,2,3,4,50000,6,7])
x.add_row([1,2,3,4,5,6000,7])

print(x.github_markdown())
```

Output:
```
1000 |  2  | 3 | 4 |     5 |    6 | 7
:----|:---:|--:|--:|------:|-----:|--:
1    |  2  | 3 | 4 |     5 |    6 | 7
1    | 200 | 3 | 4 |     5 |    6 | 7
1    |  2  | 3 | 4 | 50000 |    6 | 7
1    |  2  | 3 | 4 |     5 | 6000 | 7
```

Rendered on GitHub:

1000 |  2  | 3 | 4 |     5 |    6 | 7
:----|:---:|--:|--:|------:|-----:|--:
1    |  2  | 3 | 4 |     5 |    6 | 7
1    | 200 | 3 | 4 |     5 |    6 | 7
1    |  2  | 3 | 4 | 50000 |    6 | 7
1    |  2  | 3 | 4 |     5 | 6000 | 7

pytools/__init__.py
```diff
@@ -1489,10 +1489,12 @@ class Table:
     .. automethod:: add_row
     .. automethod:: __str__
     .. automethod:: latex
+    .. automethod:: github_markdown
     """
 
-    def __init__(self):
+    def __init__(self, alignments=['l', 'r']):
```
inducer, line 1495: Docs for `alignments`?

inducer, line 1495: I don't like that default: What if the table has more than two columns?

pytools/__init__.py
```diff
@@ -1510,6 +1512,28 @@ def __str__(self):
...
+        col_widths = [max(len(row[i]) for row in self.rows)
+                      for i in range(columns)]
+
+        alignments = self.alignments
+        # If not all alignments were specified, extend alignments with the
+        # last alignment specified:
+        alignments += self.alignments[-1] * (columns - len(self.alignments))
+
+        lines = [" | ".join([cell.center(col_width) if align == 'c'
+            else cell.ljust(col_width) if align == 'l' else cell.rjust(col_width)
```
inducer: Get all the `if`s to the beginning of the line.

pytools/__init__.py
```diff
@@ -1510,6 +1512,30 @@ def __str__(self):
...
+            else cell.rjust(col_width)
+            for cell, col_width, align in zip(row, col_widths, alignments)])
+            for row in self.rows]
+        lines[1:1] = ["|".join(
+            ":" + "-" * (col_width - 1 + (i > 0)) + ":" if align == "c"
+            else ":" + "-" * (col_width + (i > 0)) if align == "l"
+            else "-" * (col_width + (i > 0)) + ":"
+            for i, (col_width, align) in enumerate(zip(col_widths, alignments)))]
+
+        return "\n".join(lines)
```
inducer, line 1537: Also: please add a test.

pytools/__init__.py
```diff
@@ -1489,10 +1489,12 @@ class Table:
     .. automethod:: add_row
     .. automethod:: __str__
     .. automethod:: latex
+    .. automethod:: github_markdown
     """
 
-    def __init__(self):
+    def __init__(self, alignments=['l', 'r']):
         self.rows = []
+        self.alignments = alignments
```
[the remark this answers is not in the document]
matthiasdiener: Not sure. I was reluctant to change the format of the default output via `__str__`, since someone might already rely on the existing format. For `latex()` it won't make a difference, since there is no alignment.

pytools/__init__.py
matthiasdiener: Added doctest.

pytools/__init__.py
matthiasdiener: Since the ifs bind to the code before them in this case (e.g., `cell.center(col_width) if align == 'c'` belongs together) , doesn't it make sense to keep them on the same line?

pytools/__init__.py
matthiasdiener: Added.

matthiasdiener: Added in the doc - when a column has no alignment specified, it will use the last alignment in the list. In the default case shown here, all columns after the second will also have an 'r' alignment.

pytools/__init__.py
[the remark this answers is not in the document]
inducer: Could you pick the default alignments in such a way that they don't change the behavior?

pytools/__init__.py
```diff
@@ -1486,13 +1486,22 @@ def __call__(self, *args, **kwargs):
...
+        alignment.
+
     .. automethod:: add_row
     .. automethod:: __str__
     .. automethod:: latex
+    .. automethod:: github_markdown
     """
 
-    def __init__(self):
+    def __init__(self, alignments=['l', 'r']):
```
inducer, line 1502: Mutable default values are often problematic: If something does mutate them, you're in trouble.

pytools/__init__.py
```diff
@@ -1510,6 +1519,48 @@ def __str__(self):
...
+
+        .. doctest ::
+
+            >>> tbl = Table(alignments=['l', 'r'])
+            >>> tbl.add_row([1, '|'])
+            >>> tbl.add_row([10, 20])
+            >>> s = tbl.github_markdown().splitlines()
+            >>> assert s[0] == "1  | \\|"
+            >>> assert s[1] == ":--|---:"
+            >>> assert s[2] == "10 | 20"
```
inducer, line 1535: You could just print the output and have the doctest make sure it matches. (That's its job.)

pytools/__init__.py
matthiasdiener: Fixed.

pytools/__init__.py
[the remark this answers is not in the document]
matthiasdiener: Changed default alignments to `['l']`, which should match the current behavior.

pytools/__init__.py
matthiasdiener: Fixed.

matthiasdiener:
I added alignment support for __str__ as well, with the default matching the previous out.

pytools/__init__.py
```diff
@@ -1486,30 +1486,106 @@ def __call__(self, *args, **kwargs):
...
+
+        .. doctest ::
+
+            >>> tbl = Table(alignments=['l', 'r', 'l'])
+            >>> tbl.add_row([1, '|'])
+            >>> tbl.add_row([10, '20||'])
+            >>> print(tbl)
+            1  |    |
+            ---+------
+            10 | 20||
```
inducer, line 1523: Is the duplicated `|` intentional here?

pytools/__init__.py
```diff
@@ -1486,30 +1486,106 @@ def __call__(self, *args, **kwargs):
...
+
+        .. doctest ::
+
+            >>> tbl = Table(alignments=['l', 'r', 'l'])
+            >>> tbl.add_row([1, '|'])
+            >>> tbl.add_row([10, '20||'])
+            >>> print(tbl.github_markdown())
+            1  |     \|
+            :--|-------:
+            10 | 20\|\|
```
inducer, line 1560: Same here: Is the duplicated `|` intentional?

pytools/__init__.py
```diff
@@ -1486,30 +1486,106 @@ def __call__(self, *args, **kwargs):
...
             for row in self.rows]
         lines[1:1] = ["+".join("-" * (col_width + 1 + (i > 0))
             for i, col_width in enumerate(col_widths))]
 
         return "\n".join(lines)
 
+    def github_markdown(self):
+        """Returns a string representation of the table formatted as
+        GitHub-Flavored Markdown.
+        https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables
```
inducer, line 1550:
```suggestion
        `GitHub-Flavored Markdown.
        <https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables>`__
```

pytools/__init__.py
inducer: Fair enough.

pytools/__init__.py
matthiasdiener: Yes. The intention was to test that escaping the `|` works correctly, even with multiple `|` in the same table field.

pytools/__init__.py
matthiasdiener: Yes, the test input matches the one in `github_markdown` and is supposed to illustrate the difference in the output between the two functions.

inducer: OIC. I had missed the pipe characters in the input.

PR #33, opened 2020-08-26
Doc fixes

inducer:


.github/workflows/ci.yml
```diff
@@ -71,4 +71,21 @@ jobs:
...
+        steps:
+        -   uses: actions/checkout@v2
+        -
+            uses: actions/setup-python@v1
+            with:
+                python-version: '3.x'
+        -   name: "Main Script"
+            run: |
+                EXTRA_INSTALL="numpy"
+                curl -L -O -k https://gitlab.tiker.net/inducer/ci-support/raw/master/ci-support.sh
```
kaushikcfd, line 86: Can this be replaced by:
```
curl -L -O -k https://gitlab.tiker.net/inducer/ci-support/raw/master/build-docs.sh
. build-docs.sh
```
?

inducer: I'd prefer not. Sourcing `ci-support.sh` and then calling functions from it is now the preferred way.

PR #34, opened 2020-08-29
Table: add csv output

matthiasdiener:


pytools/__init__.py
```diff
@@ -1597,6 +1597,32 @@ def github_markdown(self):
 
         return "\n".join(lines)
 
+    def csv(self, dialect='excel', **fmtparams):
```
inducer, line 1600: Maybe make `fmt_params` a separate argument? Otherwise we'll have a bear of a time extending this interface later.

inducer, line 1600: Document `fmt_params` (or maybe `csv_kwargs`), e.g. by linking to `:func:\`csv.writer\``.

matthiasdiener: Fixed (if I understood your comment correctly).

matthiasdiener: Added.

inducer:
LGTM. Thanks for your contribution!

PR #37, opened 2020-09-15
Add helpers for launching MPI jobs on various platforms

majosm:
This has been generalized a little more from what you saw last week, mainly to make it easier to support functionality that exists on some platforms but not on others (like GPUs). I think it's now at a point where we can start using it and add on to it as needed.

The meshmode usage is [here](https://github.com/majosm/meshmode/blob/e43856b3aa10f23ddda0f8656befa0afdc828ed7/test/test_partition.py#L537-L561). I ended up deciding not to do the `@mpi_test` decorator (though I did get it working). I think it would just cause confusion, since there are things you can do in a normal test that you can't do in an MPI-launched test (`pytest.skip()` for example).

I tested this on my Mac, quartz, and lassen; it behaves as expected on all three.

### DIFF
--- a/pytools/mpi.py
+++ b/pytools/mpi.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+import abc
+
 
 def check_for_mpi_relaunch(argv):
     if argv[1] != "--mpi-relaunch":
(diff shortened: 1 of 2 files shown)


majosm:
How should we handle the `mpi4py` dependency?

Also, in order for the MPI tests to run, the CI will need to set the `MPI_EXECUTOR_TYPE` environment variable (presumably to `basic`).

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
     check_call(["mpirun", "-np", str(ranks),
         sys.executable, py_script, "--mpi-relaunch", callable_and_args],
         env=newenv)
+
+
+# {{{ MPI executors
```
inducer, line 36: Could you link this into the documentation? (There might not be an MPI section yet, please create one. :)

test/test_mpi.py
```diff
@@ -0,0 +1,116 @@
...
+        MPIExecError,
+        MPIExecutorParams,
+        MPIExecutorParamError)
+
+import pytest
+import os
+from functools import partial
+
+
+def get_test_mpi_executor():
```
inducer, line 15: Do these tests run as part of CI? If not I'd like them to.

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
...
+
+
+class MPIExecError(RuntimeError):
+    def __init__(self, exit_code):
+        self.exit_code = exit_code
+        super().__init__(f"MPI execution failed with exit code {exit_code}.")
+
+
+class MPIExecutor(metaclass=abc.ABCMeta):
+    """Base class for a general MPI launcher."""
```
inducer, line 78: Add
```
.. automethod:: xxxy
```
to the the class docstring for each method.

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
...
+        """
+        exit_code = self.execute(code_string, exec_params)
+        if exit_code != 0:
+            raise MPIExecError(exit_code)
+
+    def call(self, func, exec_params=None):
+        """Calls *func* with MPI. Note: *func* must be picklable."""
+        import pickle
+        calling_code = ('import sys; import pickle; pickle.loads(bytes.fromhex("'
+                    + pickle.dumps(func).hex() + '"))()')
```
inducer, line 111: `.hex()` is not an efficient encoding. Base64 is better.

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
...
+    def check_execute(self, code_string, exec_params=None):
+        """
+        Executes Python code stored in *code_string* with MPI and raises an instance
+        of :class:`MPIExecError` if the execution fails.
+        """
+        exit_code = self.execute(code_string, exec_params)
+        if exit_code != 0:
+            raise MPIExecError(exit_code)
+
+    def call(self, func, exec_params=None):
```
inducer, line 107: It might be useful to allow arguments here. Not every user might know about `functools.partial`.

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
...
+        """Calls *func* with MPI. Note: *func* must be picklable."""
```
inducer, line 108: Document `exec_params`. (Here and elsewhere, by reference "See `:meth:\`...\`` for *exec_params*" OK.)

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
...
+
+    :arg executor_type_name: The executor type name. Can be one of `'basic'`,
+        `'slurm'`, or `'lclsf'`.
+    """
+    type_name_map = {
+        "basic": BasicMPIExecutor,
+        "slurm": SlurmMPIExecutor,
+        "lclsf": LCLSFMPIExecutor
+    }
+    return type_name_map[executor_type_name]()
```
inducer, line 210: Would it make sense to create some rudimentary auto-detection facility (which can be overridden by a documented env var)?

pytools/mpi.py
```diff
@@ -29,3 +31,184 @@ def run_with_mpi_ranks(py_script, ranks, callable_, args=(), kwargs=None):
     check_call(["mpirun", "-np", str(ranks),
         sys.executable, py_script, "--mpi-relaunch", callable_and_args],
         env=newenv)
+
+
+# {{{ MPI executors
+
+class MPIExecutorParams:
```
inducer, line 38: Dataclass maybe?

inducer (commented):
Thanks for working on this. A few comments below.

inducer:
> Also, in order for the MPI tests to run, the CI will need to set the `MPI_EXECUTOR_TYPE` environment variable (presumably to `basic`).

Sure. You can hack it in yourself in `.github/workflows/ci.yml` and `.gitlab-ci.yml`. Please submit a parallel MR to https://gitlab.tiker.net/inducer/pytools/ (added you), since you'll need to hack its CI config, too.

inducer:
> How should we handle the `mpi4py` dependency?

For the CI? No big deal IMO. Just install it. But pytools itself should not declare a hard dependency on mpi4py.

Also, maybe this stuff wants to be a separate package. What do you think?

pytools/mpi.py
majosm: Is there a convention for passing in the argument list? As a `dict`? Or something else?

majosm:
> Also, maybe this stuff wants to be a separate package. What do you think?

I made an attempt: [link](https://github.com/majosm/squirm)

PR #45, opened 2020-10-15
Annotate the types for pytools.product

kaushikcfd:


inducer:
Thx!

pytools/__init__.py
```diff
@@ -1049,7 +1049,7 @@ def partition2(iterable):
     return part_true, part_false
 
 
-def product(iterable):
+def product(iterable: Iterable[Any]) -> Any:
```
inducer, line 1052: I wonder if a type variable (instead of `Any`) would have been a better approach here.

PR #49, opened 2020-10-22
Add a user agent

alexfikl:
By default, it seems like python uses `Python-urllib/<version>`, which I guess some servers block.


inducer:
Thx!

PR #52, opened 2020-11-26
Escape backslashes in Table documentation

slavoutich:
Sequence of symbols `\|` is an invalid escape sequence in Python and may cause syntax warning or syntax error in certain Python versions. An example of that is a failing test here:
https://gitlab.com/quantumsim/quantumsim/-/jobs/869297979
This PR escapes these backslashes in docstring, that fixes the error above, as it happens in this pipeline:
https://gitlab.com/quantumsim/quantumsim/-/jobs/875855857

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1614,9 +1614,9 @@ class Table:
             >>> tbl.add_row([1, '|'])
             >>> tbl.add_row([10, '20||'])
             >>> print(tbl.github_markdown())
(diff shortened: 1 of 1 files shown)


inducer:
How about #53 instead?

slavoutich:
It's a matter of your taste, both fix the issue :)

inducer:
Sounds good, merged #53. Thanks for reporting the issue.

PR #57, opened 2020-12-10
Add support for parsing dotted names into tags

kaushikcfd:


pytools/tag.py
```diff
@@ -130,4 +135,148 @@ class UniqueTag(Tag):
...
+
+    tag_class: module "." name    -> map_tag_class
+
+    module: name                  -> map_top_level_module
+          | module "." name       -> map_nested_module
+
+    name: CNAME                   -> map_name
+
+    % if shortcuts:
+    SHORTCUT: (${" | ".join('"' + name + '"' for name in shortcuts)})
```
inducer, line 173: I think the grammar should be the same no matter what shortcuts are defined. If that makes the shortcuts slightly more cumbersome to type, so be it.

setup.py
```diff
@@ -41,7 +41,9 @@
           "decorator>=3.2.0",
           "appdirs>=1.4.0",
           "numpy>=1.6.0",
-          "dataclasses>=0.7;python_version<='3.6'"
+          "dataclasses>=0.7;python_version<='3.6'",
+          "Mako",
+          "lark-parser",
```
inducer, line 46: I'd like to avoid the dependency on Mako, and I'd like the lark dependency to be an "extra" (e.g. `"tag_parsing"?).

test/testlib.py
```diff
@@ -0,0 +1,26 @@
+from pytools.tag import Tag, UniqueTag
```
inducer, line 1: Rename to `testlib_tags`?

inducer (commented):
Thanks for working on this. Some initial high-level feedback below.

test/testlib.py
kaushikcfd: Done.

setup.py
kaushikcfd: Done. For installing pytools with tag parsing, now one would do: `pip install pytools[tag_parsing]`.

pytools/tag.py
kaushikcfd: Agreed, so the shortcuts are now prepended with a period (`"."`). That still allows them to be a convenient shortcut and also lets us having a well-defined grammar.

pytools/tag.py
```diff
@@ -257,4 +262,152 @@ def without_tags(self: T_co,
...
+        return self.shortcuts[name]
+
+
+def parse_tag(tag_text, shortcuts={}):
+    """
+    Parses a :class:`Tag` from a provided dotted name.
+    """
+    import inspect
+    caller_globals = inspect.currentframe().f_back.f_globals
+    parser = Lark(TAG_GRAMMAR, start="tag")
```
inducer, line 405: How costly is that parser construction? Maybe we could `@memoize` it?

inducer, line 405: Does this use LALR or Earley? (I ask because we should IMO shoot for LALR as much as possible, or at least not prevent its use for a simple grammar like this.)

pytools/tag.py
```diff
@@ -1,5 +1,8 @@
 from dataclasses import dataclass
 from typing import Tuple, Any, FrozenSet, Union, Iterable, TypeVar
+import importlib
+
+from lark import Lark, Transformer
```
inducer, line 5: Could you make it so that `pytools.tag` is still importable without `lark`?

.gitlab-ci.yml
```diff
@@ -7,7 +7,7 @@ Pytest:
     # AK, 2020-12-13
     rm pytools/log.py
 
-    export EXTRA_INSTALL="numpy"
+    export EXTRA_INSTALL="numpy lark-parser"
```
inducer, line 10: Could you push a sister MR to Gitlab (add the link to the PR description) to ensure that passes, too?

inducer (commented):
Thanks! Some initial comments below.

pytools/tag.py
```diff
@@ -257,4 +262,152 @@ def without_tags(self: T_co,
...
+    SHORTCUT: "." ("_"|LETTER) ("_"|LETTER|DIGIT|".")*
+
+    %import common.INT
+    %import common.ESCAPED_STRING
+    %import common.DIGIT
+    %import common.LETTER
+    %import common.CNAME
+    %import common.WS
+    %ignore WS
+"""
```
inducer, line 304: This should probably be a separate `.lark` file so that it's importable elsewhere.

https://lark-parser.readthedocs.io/en/latest/grammar.html#import

inducer: (Unless it's importable this way? The docs didn't make it sound that way.)

inducer: If it becomes a `.lark` file, make sure to also add that to `MANIFEST.in`.

inducer, line 304: Looked over the grammar, looks OK. (Please leave this for me to resolve at final review.)

pytools/tag.py
kaushikcfd: Yep, see ff625d904bf91752d6c0e5dc1b4679e652c150a4.

pytools/tag.py
kaushikcfd: It was using Earley earlier. Now it uses LALR. Bonus: now we also get faster first-time construction as lark supports on-disk parser object caching for LALR type parsers. (ff625d904bf91752d6c0e5dc1b4679e652c150a4)

kaushikcfd: Done in ff625d904bf91752d6c0e5dc1b4679e652c150a4.

pytools/tag.py
kaushikcfd: Yup, done. See 6583c32b3c74745803243c0e4df0d2d96cf32a54.

.gitlab-ci.yml
kaushikcfd: https://gitlab.tiker.net/inducer/pytools/-/merge_requests/52

kaushikcfd:
Probably I should re-write this as `pytools.resolve_name` already contains most of the logic.

inducer:

inducer:
> Probably I should rewrite this as `pytools.resolve_name` already contains most of the logic.

I actually don't think so. We'll be able to reuse this as we use more lark. Sorry I've been slow to review. Otherwise ready for another look from your end?

PR #67, opened 2021-03-10
memoize_method for frozen classes

alexfikl:
Went ahead and tried using `object.__setattr__` in `memoize_method` and `keyed_memoize_method`. What do you think?

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -675,8 +675,9 @@ def memoize_on_first_arg(function, cache_dict_name=None):
     """
 
     if cache_dict_name is None:
-        cache_dict_name = intern("_memoize_dic_"
-                + function.__module__ + function.__name__)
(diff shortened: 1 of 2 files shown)


pytools/__init__.py
```diff
@@ -675,8 +675,9 @@ def memoize_on_first_arg(function, cache_dict_name=None):
     """
 
     if cache_dict_name is None:
-        cache_dict_name = intern("_memoize_dic_"
-                + function.__module__ + function.__name__)
+        cache_dict_name = intern(
+                "_memoize_dic_{function.__module__}{function.__name__}"
```
kaushikcfd, line 679: Should this be an f-string?

alexfikl: It really should! Thanks!

Should be fixed in latest version.

inducer:
LGTM. Thanks!

PR #69, opened 2021-03-20
Avoid floating point zero in EOCRecorder

alexfikl:


pytools/convergence.py
```diff
@@ -30,6 +30,10 @@ def estimate_order_of_convergence(self, gliding_mean=None):
         abscissae = np.array([a for a, e in self.history])
         errors = np.array([e for a, e in self.history])
 
+        # NOTE: in case any of the errors are exactly 0.0, which
+        # can give NaNs in `estimate_order_of_convergence`
+        errors += np.finfo(errors.dtype).eps
```
inducer, line 35: Maybe scale this by the max of the errors? (Or 1 if they're all zero?)

alexfikl: Done in c78dd0c.

inducer:
Thanks!

PR #76, opened 2021-04-04
Fix some typos after decorator removal

alexfikl:
Added some tests too.

### DIFF
--- a/pytools/obj_array.py
+++ b/pytools/obj_array.py
@@ -216,6 +216,8 @@ def obj_array_vectorize_n_args(f, *args):
 def obj_array_vectorized_n_args(f):
     # Unfortunately, this can't use partial(), as the callable returned by it
     # will not be turned into a bound method upon attribute access.
+    # This may happen here, because the decorator *could* be used
+    # on methods, since it can "look past" the leading `self` argument.
     # Only exactly function objects receive this treatment.
     #
     # Spec link:
--- a/pytools/obj_array.py
+++ b/pytools/obj_array.py
@@ -224,7 +224,7 @@ def obj_array_vectorized_n_args(f):
     # > Also notice that this transformation only happens for user-defined functions;
     # > other callable objects (and all non-callable objects) are retrieved
     # > without transformation.
-    
+
     def wrapper(*args):
         return obj_array_vectorize_n_args(f, *args)
 
--- a/pytools/obj_array.py
+++ b/pytools/obj_array.py
@@ -214,10 +214,7 @@ def obj_array_vectorize_n_args(f, *args):
 
 
 def obj_array_vectorized_n_args(f):
-    def wrapper(*args, **kwargs):
-        if len(kwargs):
-            raise NotImplementedError("kwargs are not supported")
-
+    def wrapper(*args):
         return obj_array_vectorize_n_args(f, *args)
 
     update_wrapper(wrapper, f)
--- a/pytools/obj_array.py
+++ b/pytools/obj_array.py
@@ -214,7 +214,12 @@ def obj_array_vectorize_n_args(f, *args):
 
 
 def obj_array_vectorized_n_args(f):
-    wrapper = partial(obj_array_vectorize_n_args, f)
+    def wrapper(*args, **kwargs):
+        if len(kwargs):
+            raise NotImplementedError("kwargs are not supported")
+
+        return obj_array_vectorize_n_args(f, *args)
+
     update_wrapper(wrapper, f)
     return wrapper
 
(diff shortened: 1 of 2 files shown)


test/test_pytools.py
```diff
@@ -362,6 +362,81 @@ def test_make_obj_array_iteration():
...
+    @obj.obj_array_vectorized_n_args
+    def vectorized_two_add_one(x, y):
+        assert x.dtype.char != "O" and y.dtype.char != "O"
+        return x * y + c
+
+    class Adder:
+        def __init__(self, c):
+            self.c = c
+
+        @obj.obj_array_vectorized_n_args
```
alexfikl, line 397: Adding this as a decorator on a class doesn't actually work anymore.

From looking at the code in `decorator`, they do some fancy stuff with `inspect` to get `self` back into the arguments. I'll look into fixing that too.

alexfikl: 236e746?

inducer: Wow. I just spent a long time understanding what's going on here, as evidenced by the blurb I added above. TIL about a less-than-pretty corner of Python. Long story short: What you did LGTM. :)

inducer (commented):
Thanks for working on this! I'll roll a new release once this is in. meshmode is probably also affected: https://github.com/inducer/meshmode/blob/ffdea6a19fc609be295b551d167f995df7700be0/meshmode/dof_array.py#L421-L424

test/test_pytools.py
alexfikl: I'm still not sure I understand it properly!

pytools/obj_array.py
```diff
@@ -214,9 +214,14 @@ def obj_array_vectorize_n_args(f, *args):
 
 
 def obj_array_vectorized_n_args(f):
-    wrapper = partial(obj_array_vectorize_n_args, f)
+    def wrapper(*args, **kwargs):
+        if len(kwargs):
+            raise NotImplementedError("kwargs are not supported")
+
```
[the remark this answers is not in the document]
alexfikl: I just thought it was nicer to complain if someone tried to add keyword arguments, but this is probably just as clear.

alexfikl:
> Thanks for working on this! I'll roll a new release once this is in. meshmode is probably also affected: [inducer/meshmode@`ffdea6a`/meshmode/dof_array.py#L421-L424](https://github.com/inducer/meshmode/blob/ffdea6a19fc609be295b551d167f995df7700be0/meshmode/dof_array.py#L421-L424)

Yep, I'll open a PR there too!

pytools/obj_array.py
```diff
@@ -214,9 +214,11 @@ def obj_array_vectorize_n_args(f, *args):
 
 
 def obj_array_vectorized_n_args(f):
-    wrapper = partial(obj_array_vectorize_n_args, f)
+    def wrapper(*args):
```
inducer, line 217:
```suggestion
    # Unfortunately, this can't use partial(), as the callable returned by it
    # will not be turned into a bound method upon attribute access.
    # Only exactly function objects receive this treatment.
    #
    # Spec link:
    # https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy
    # (under "Instance Methods", quote as of Py3.9.4)
    # > Also notice that this transformation only happens for user-defined functions;
    # > other callable objects (and all non-callable objects) are retrieved
    # > without transformation.
    
    def wrapper(*args):
```

inducer:
I had typed out a different version of the explanation in 61b3aeb earlier, but somehow that never made it into the review. Not sure why. At any rate, that's the explanation I came up with.

> Yep, I'll open a PR there too!

Awesome, thanks! Maybe include the explanation there, too, just to limit confusion. (I'm sure I'll have forgotten what's up in a few months.)

pytools/obj_array.py
```diff
@@ -214,9 +214,22 @@ def obj_array_vectorize_n_args(f, *args):
 
 
 def obj_array_vectorized_n_args(f):
-    wrapper = partial(obj_array_vectorize_n_args, f)
+    # Unfortunately, this can't use partial(), as the callable returned by it
+    # will not be turned into a bound method upon attribute access.
```
inducer, line 218:
```suggestion
    # will not be turned into a bound method upon attribute access.
    # This may happen here, because the decorator *could* be used
    # on methods, since it can "look past" the leading `self` argument.
```

inducer:
Explanation expanded slightly in ea08b37.

inducer:
Released as 2021.2.3.

PR #77, opened 2021-04-06
Stringify ProcessTimer

alexfikl:
Can now do
```
with ProcessTimer() as p:
	# do work
print(p)
```
Also removed support for python<3.3, since pytools advertises >3.6.


inducer:
Thanks!

PR #80, opened 2021-04-15
Steps towards more symbolic dict keys in memoize_on_first_arg, memoize_method

inducer:


pytools/__init__.py
```diff
@@ -678,33 +674,34 @@ def memoize_on_first_arg(function, cache_dict_name=None):
     """
 
     if cache_dict_name is None:
-        cache_dict_name = intern(
-                f"_memoize_dic_{function.__module__}{function.__name__}"
-                )
+        cache_dict_name = (memoize_on_first_arg, function)
```
alexfikl, line 677: Does this work if `function` is nested (i.e. not in the global namespace)? It should keep changing its hash/id in that case, right?

inducer: Yeah, nested functions are a difficult case, mainly because it's hard to get stable identifiers for them. My understanding is that, at least, it would fail "safe" (separate caches for each function) using this code.

IMO, `@memoize_in` is more appropriate for nested functions.

pytools/__init__.py
```diff
@@ -678,33 +674,34 @@ def memoize_on_first_arg(function, cache_dict_name=None):
...
+            object.__setattr__(obj, "_memoize_on_first_arg_dict", {key: result})
             return result
         else:
-            getattr(obj, cache_dict_name)[key] = result
+            obj._memoize_on_first_arg_dict[key] = result
             return result
 
     def clear_cache(obj):
-        object.__delattr__(obj, cache_dict_name)
+        del obj._memoize_on_first_arg_dict
```
inducer, line 704:
```suggestion
        object.__delattr__("_memoize_on_first_arg_dict")
```

pytools/__init__.py
```diff
@@ -678,33 +674,34 @@ def memoize_on_first_arg(function, cache_dict_name=None):
...
+            object.__setattr__(obj, "_memoize_on_first_arg_dict", {key: result})
             return result
         else:
-            getattr(obj, cache_dict_name)[key] = result
+            obj._memoize_on_first_arg_dict[key] = result
             return result
 
     def clear_cache(obj):
-        object.__delattr__(obj, cache_dict_name)
+        object.__delattr__("_memoize_on_first_arg_dict")
```
inducer, line 704:
```suggestion
        object.__delattr__(self, "_memoize_on_first_arg_dict")
```

pytools/__init__.py
```diff
@@ -678,33 +674,34 @@ def memoize_on_first_arg(function, cache_dict_name=None):
...
+            object.__setattr__(obj, "_memoize_on_first_arg_dict", {key: result})
             return result
         else:
-            getattr(obj, cache_dict_name)[key] = result
+            obj._memoize_on_first_arg_dict[key] = result
             return result
 
     def clear_cache(obj):
-        object.__delattr__(obj, cache_dict_name)
+        object.__delattr__(self, "_memoize_on_first_arg_dict")
```
inducer, line 704:
```suggestion
        object.__delattr__(obj, "_memoize_on_first_arg_dict")
```

inducer:
(I added loopy downstream CI because I happen to know that this change breaks loopy.)

pytools/__init__.py
alexfikl: Maybe add a small note to the docs that it doesn't work for nested?

alexfikl: Besides that, it looks good to me! :+1:

inducer:
> (I added loopy downstream CI because I happen to know that this change breaks loopy.)

Huh, TIL it doesn't. *scratches head* All the better. :)

pytools/__init__.py
inducer: 219e111

inducer:
Turns out the downstream CIs weren't actually using this code:
```
Requirement already satisfied: pytools from git+file:/home/runner/work/pytools/pytools#egg=pytools in ./.miniforge3/envs/testing/lib/python3.8/site-packages (from -r requirements.txt (line 1)) (2021.2.3)
```

 #82

inducer:
There we go. :) Downstream CI is not such a bad thing... when it works.

PR #82, opened 2021-04-17
Fix downstream CI: force use of newly installed package

inducer:


inducer:
Hmm... looks like that worked.

PR #84, opened 2021-04-22
Un-methodize check_tag_uniqueness and normalize_tags

inducer:
cc @thomasgibson @nchristensen

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -49,6 +49,13 @@ Supporting Functionality
 .. autoclass:: DottedName
 .. autoclass:: NonUniqueTagError
 
+
+Internal stuff that is only here because the documentation tool wants it
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. class:: T_co
+
+    A covariant type variable used in, e.g. :class:`Taggable.copy`.
 """
 
 #  }}}
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -236,6 +236,9 @@ class Taggable:
     """
 
     def __init__(self, tags: TagsType = frozenset()):
+        assert isinstance(tags, FrozenSet)
+        assert all(isinstance(tag, Tag) for tag in tags)
+        check_tag_uniqueness(tags)
         self.tags = tags
 
     def copy(self: T_co, **kwargs: Any) -> T_co:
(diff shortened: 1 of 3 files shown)


pytools/tag.py
```diff
@@ -195,34 +236,10 @@ class Taggable:
     """
 
     def __init__(self, tags: TagsType = frozenset()):
-        # For performance we assert rather than
-        # normalize the input.
         assert isinstance(tags, FrozenSet)
         assert all(isinstance(tag, Tag) for tag in tags)
+        check_tag_uniqueness(tags)
```
thomasgibson, line 241: @inducer I had to add these back in to get the test suite to pass. This is because you perform error handling in the unit tests: https://github.com/inducer/pytools/blob/main/test/test_pytools.py#L482-L496

thomasgibson (commented):
Just a comment on error handling in the constructor. We may want to leave the assertion checks in the `Taggable` constructor or update the unit tests

pytools/tag.py
inducer: I'd say let's update the unit tests and document the policy on where input checking is to be done (i.e. not the constructor).

thomasgibson: @inducer I've gone ahead and updated the documentation and unit tests. I figured it's best to use `check_tag_uniqueness` for input-checking, and so I've documented that in the `Taggable` constructor.

I also modified the `check_tag_uniqueness` slightly to include an `isinstance(tag, Tag)` (since we already loop through them, might as well check they make sense). I removed the corresponding line from `normalize_tags` since it's always used internally and passed through `check_tag_uniqueness` in the relevant routines. Let me know if you think this is okay!

thomasgibson:
I think we might be running into a Sphinx bug. Docs are being built with Sphinx v 3.5.4 (according to the CI log), and check out this issue: https://github.com/sphinx-doc/sphinx/issues/8127 --- I think this fix will only be available in v 4+ releases

inducer:
How about 69e8148?

inducer:
LGTM.

pytools/tag.py
```diff
@@ -36,7 +36,9 @@
 
 Tag Interface
 ---------------
+.. <!--``normalize_tags`` undocumented for now. (Not ready to commit.)-->
```
inducer, line 39: Out of curiosity: How come this got wrapped in an HTML comment?

thomasgibson: This was just to prevent this from being rendered in the generated HTML while keeping this here as a comment

inducer: I think 57678bf suffices.

inducer:
Thanks!

inducer:
Released: https://pypi.org/project/pytools/2021.2.4/#history

PR #94, opened 2021-06-14
PicklableFunction: overwrite __call__ instead of call in __call__

inducer:


inducer:
@matthiasdiener This provides a speed increase that's worth having if we can add justifiable conditionals for Python versions. It seems to work on Py3.9, but as the CI shows, not on Py3.7. Before moving forward with it, I was trying to find someplace that mentions that this behavior changed in

- https://docs.python.org/3/whatsnew/3.9.html
- https://docs.python.org/3/whatsnew/3.8.html

I didn't find anything though. Do you see something?

inducer:
Weird. Am I seeing a ghost? This passes on dunkel's Python 3.9.

pytools/py_codegen.py
```diff
@@ -144,10 +145,14 @@ def __init__(self, module, name):
...
+        if sys.version_info < (3, 9):
+            self.func = module.mod_globals[name]
+        else:
+            self.__call__ = module.mod_globals[name]
 
-    def __call__(self, *args, **kwargs):
-        return self.func(*args, **kwargs)
+    if sys.version_info < (3, 8):
+        def __call__(self, *args, **kwargs):  # pylint: disable=method-hidden
+            return self.func(*args, **kwargs)
```
alexfikl, line 155:
```suggestion
if sys.version_info < (3, 8):
    def __call__(self, *args, **kwargs):  # pylint: disable=method-hidden
        return self.func(*args, **kwargs)
```
? Otherwise that `__call__` doesn't seem to be doing anything?

inducer: No, that's intentional. The `__call__` method definition should only run on Py3.7 and below. (Or at least that's the thought...)

alexfikl: Yes, but it's only defined in the `_initialize`, not at the class level, right? (that's why I unindented the whole block)

Might be looking at it weird through the diff..

alexfikl: Yep, looked at it weird, nevermind!

inducer:
Hmm... nvm. I had my virtualenv tangled, hence the weird results...

inducer:
Sorry about the confusion!

pytools/py_codegen.py
```diff
@@ -144,10 +145,14 @@ def __init__(self, module, name):
     def _initialize(self, module, name):
         self.module = module
         self.name = name
-        self.func = module.mod_globals[name]
+        if sys.version_info < (3, 9):
+            self.func = module.mod_globals[name]
+        else:
+            self.__call__ = module.mod_globals[name]
```
alexfikl, line 151: According to
https://docs.python.org/3/reference/datamodel.html#special-lookup
This shouldn't really work?

PR #97, opened 2021-07-16
make Taggable.copy an abstractmethod

kaushikcfd:
This should have helped catch some errors in [Pytato](https://github.com/inducer/pytato), as the taggable-type didn't have a `.copy` defined.

Draft because:
- [x] Needs https://github.com/inducer/pytato/pull/125 (?)
- [x]  Needs https://github.com/inducer/loopy/pull/467

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -27,6 +27,7 @@ Internal stuff that is only here because the documentation tool wants it
 from dataclasses import dataclass
 from typing import Tuple, Set, Any, FrozenSet, Union, Iterable, TypeVar
 from pytools import memoize
+from abc import ABC, abstractmethod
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks, this LGTM. I'm converting this to draft as merging it would break pytato. I'm guessing https://github.com/inducer/pytato/pull/125 would be required to prevent that; I'll change the description to mention that.

kaushikcfd:
(Closed + re-opened to re-trigger the tests to find out if there are more downstream users that violate this restriction)

inducer:
(Added you to the repo so you can just rerun.)

inducer:
Alright, looks like this is ready. In it goes. Thanks!

PR #99, opened 2021-08-16
Allow choosing table type in EOCRecorder

alexfikl:
Set the default to `markdown` instead of the previous `ascii` since the tables look pretty much the same and this is easier to copy paste in here.


inducer:
Thanks!

PR #103, opened 2021-10-11
Allow passing threshold to log_process decorator

alexfikl:


pytools/__init__.py
```diff
@@ -2526,15 +2526,17 @@ class log_process:  # noqa: N801
     to the wrapped function.
     """
```
inducer, line 2527:
```suggestion
    .. automethod:: __init__
    .. automethod:: __call__
    """
```

PR #104, opened 2021-10-12
pyupgrade more

alexfikl:
Just remembered I had this lying around from a few weeks back. 

A good chunk is automated by `pyupgrade`, but it has some manual f-strings changes, but hopefully no typos :crossed_fingers:.


inducer:
LGTM, thanks!

PR #105, opened 2021-10-29
Improve EOCRecorder docs and checks

alexfikl:
* Adds some types
* Adds methods to docs
* Adds some asserts in `add_data_point` to make sure that it doesn't get weird non-numeric data

### DIFF
--- a/pytools/convergence.py
+++ b/pytools/convergence.py
@@ -1,3 +1,5 @@
+from typing import List, Optional, Tuple
+
 import numpy as np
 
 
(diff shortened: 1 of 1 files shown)


pytools/convergence.py
```diff
@@ -20,13 +22,32 @@ def estimate_order_of_convergence(abscissae, errors):
...
     def __init__(self):
-        self.history = []
+        self.history: List[Tuple[float, float]] = []
+
+    def add_data_point(self, abscissa: float, error: float) -> None:
+        from numbers import Number
+        assert isinstance(abscissa, Number), \
+                f"'abscissa' is not a number: {type(abscissa).__name__}"
+        assert isinstance(error, Number), \
+                f"'error' is not a number: {type(error).__name__}"
```
inducer, line 44: Input checking should be `if` not `assert`.

inducer (commented):
Thanks! LGTM other than the `assert` thing.

pytools/convergence.py
alexfikl: Done in new version 3bd3f56.

inducer:
Thx!

inducer:
This may have been a bit overzealous. It extensively breaks meshmode's tests: https://github.com/inducer/pyopencl/runs/4058188593?check_suite_focus=true

inducer:
#109

PR #108, opened 2021-10-31
Fix check in EOCRecorder

alexfikl:
Sorry about the breakage! This seems to fix it locally.


inducer:
Thanks! That was quick. I like yours better than mine. :)

PR #110, opened 2021-10-31
Add more tests to EOCRecorder

alexfikl:
Added some more tests to check the inputs to `add_data_point`. Hopefully catch the breakage earlier next time!


inducer:
Thanks!

PR #112, opened 2021-11-01
Prefer platformdirs to appdirs if available

alexfikl:
Seems to be the prefered maintained alternative these days (see https://github.com/psf/black/pull/2375). Not quite sure what do to in `setup.py` though. Just replace `appdirs` with `plaformdirs` in `install_requires`?

For context, for some reason `appdirs` disappeared from my venv on porter (likely due to `--system-site-packages`) and `platformdirs` was there.


inducer:
LGTM, thanks!

PR #114, opened 2021-12-17
Remove a spurious warning.

FreddieWitherden:
This warning has been confusing our users for a while as when our main process dies due to a SEGFAULT this warning gets printed, leading to users believing this is the issue.  As is, the warning does not add much utility, so this PR simply removes it and has the process terminate gracefully.


inducer:
Thanks! I can see the use case for this change, and I think there's a low risk of confusion in the other direction (poorly understood exits of the "preforked" process), so this LGTM.

PR #116, opened 2022-01-24
Add table merging

alexfikl:
This adds a couple of things:
* type annotations to `Table`
* adds some checks when adding rows to the table.
* removes some duplication in getting alignements and column widths in the table stringifiers.
* adds a `merge_tables` that takes a bunch of tables and returns a single table with all the columns concatenated
* uses that to implement a `stringify_eocs` that merges a bunch of `EOCRecorders` into a single table.

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -103,6 +103,7 @@ Formatting
 ----------
 
 .. autoclass:: Table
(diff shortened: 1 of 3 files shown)


inducer:
Thanks!

PR #120, opened 2022-02-21
Taggable: Do not force `copy` for tag setting

inducer:
Based on https://github.com/inducer/arraycontext/pull/147#issuecomment-1046368865.

It's not safe to assume `copy` is the right verb for tag application, it's taken e.g. for `pyopencl.array.Array`. Leave more of the actual detail to the subclass.

- [x] Needs https://github.com/inducer/pytato/pull/262

cc @alexfikl @kaushikcfd

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -21,8 +21,7 @@ Internal stuff that is only here because the documentation tool wants it
 
 .. class:: T_co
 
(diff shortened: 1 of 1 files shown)


inducer:
Not hearing loud objections, been running it locally. I'll proceed with this.

PR #124, opened 2022-04-06
Fix type names in `pytools.tag`

inducer:
Upon reviewing https://github.com/inducer/arraycontext/pull/154, I found myself very much not loving the type names I was looking at. @alexfikl, could you take a look if this seems sane? (I'll do a pytools release once some version of this is in.)

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -1,5 +1,4 @@
 """
-
 Tag Interface
 ---------------
 .. ``normalize_tags`` undocumented for now. (Not ready to commit.)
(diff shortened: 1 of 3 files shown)


pytools/tag.py
```diff
@@ -162,9 +162,9 @@ class UniqueTag(Tag):
 # }}}
 
 
-TagsType = FrozenSet[Tag]
-TagOrIterableType = Union[Iterable[Tag], Tag, None]
-T_co = TypeVar("T_co", bound="Taggable")
+TagSet = FrozenSet[Tag]
```
alexfikl, line 165: Is this required? Writing `FrozenSet[Tag]` seems clearer and not that long.

alexfikl (approved):
Definitely better!

pytools/tag.py
```diff
@@ -162,9 +162,9 @@ class UniqueTag(Tag):
...
+ToTagSetConvertible = Union[Iterable[Tag], Tag, None]
+Taggable_co = TypeVar("Taggable_co", bound="Taggable")
```
alexfikl, line 167: Should this have a `TypeVar(..., covariant=True)`?

pytools/tag.py
inducer: Good point. Dropped.

pytools/tag.py
inducer: Upon further thought, what these IMO really should be is the `Self` type from Py 3.11.

inducer: (I've dropped the type variable, with deprecation.)

alexfikl: I didn't look before how it's used, but yeah, that makes sense!

inducer:
Thanks for taking a look!

cc @nchristensen

PR #125, opened 2022-04-14
Use math.perm and math.comb if available

alexfikl:
* Also added some more checks to the custom `perm` to match what the docs for `math.perm` say.
* Deprecated `pytools.factorial`.


pytools/__init__.py
```diff
@@ -267,43 +271,72 @@ def delta(x, y):
...
+            raise TypeError(f"'{type(n).__name__}' object cannot be interpreted "
+                            "as an integer")
+
+        result = 1
+        while k:
+            result *= n
+            n -= 1
+            k -= 1
+
+        return result
```
inducer, line 327: Maybe deprecate `perm` and `comb`, too, directing folks to `math`? I anticipate that we'll go to 3.8+ this year.

pytools/__init__.py
```diff
@@ -267,43 +271,72 @@ def delta(x, y):
...
-    choices.
-    """
-    result = 1
-    assert k > 0
-    while k:
-        result *= n
-        n -= 1
-        k -= 1
+try:
+    # NOTE: only available in python > 3.8
```
inducer, line 290:
```suggestion
    # NOTE: only available in python >= 3.8
```

pytools/__init__.py
```diff
@@ -267,43 +271,72 @@ def delta(x, y):
...
+        while k:
+            result *= n
+            n -= 1
+            k -= 1
+
+        return result
+
+
+try:
+    # NOTE: only available in python > 3.8
```
inducer, line 331:
```suggestion
    # NOTE: only available in python >= 3.8
```

inducer (commented):
Thanks! Just a few things below.

pytools/__init__.py
alexfikl: How does this look?
 https://github.com/inducer/pytools/compare/36dfffae49cf9dc8eb7446c98f5cd635bfd6f231..501b7a0f5178137e8fad1f4a0fc02f67d416d9ce

pytools/__init__.py
```diff
@@ -267,43 +271,83 @@ def delta(x, y):
...
+        return result
 
-    return result
+    def perm(n: SupportsIndex, k: Optional[SupportsIndex] = None) -> int:
+        """
+        :returns: :math:`P(n, k)`, the number of permutations of length :math:`k`
+            drawn from :math:`n` choices.
+        """
+        from warnings import warn
+        warn("This function is deprecated and will go away in 2022. "
```
inducer, line 307:
```suggestion
        warn("This function is deprecated and will go away in 2023. "
```

pytools/__init__.py
```diff
@@ -267,43 +271,83 @@ def delta(x, y):
...
+    # NOTE: only available in python >= 3.8
+    comb = math.comb
+except AttributeError:
+    def comb(n: SupportsIndex, k: SupportsIndex) -> int:
+        """
+        :returns: :math:`C(n, k)`, the number of combinations (subsets)
+            of length :math:`k` drawn from :math:`n` choices.
+        """
+        from warnings import warn
+        warn("This function is deprecated and will go away in 2022. "
```
inducer, line 346:
```suggestion
        warn("This function is deprecated and will go away in 2023. "
```

pytools/__init__.py
inducer, line 307: Hmm. As it stands, the deprecation warning gets shown to folks with Python 3.7 and lower, which is approximately nobody. Want to add a `__getattr__`-based warning instead?

inducer (commented):
Thanks! A few more... sorry!

pytools/__init__.py
alexfikl: @inducer Hm, yeah, I couldn't decide what's the best way to do this.
* Python >=3.8 gets a deprecation warning?
* Python <3.8 gets a deprecation warning + suggestion to update? There isn't anything to suggest to use instead here though.
* Both?

alexfikl: https://github.com/inducer/pytools/compare/501b7a0f5178137e8fad1f4a0fc02f67d416d9ce..cd515dcb143974f2eb8f1fe132298bb3b7eea515
Thoughts?

alexfikl: Ok, `mypy` seems pretty unhappy with that, but otherwise it looks good to me.

inducer:
Those look good. Thanks!

PR #131, opened 2022-04-26
Tighten typing_extensions version requirement

nchristensen:
Needs a version that includes the `Self` type.


inducer:
Good catch.

- Added here: https://github.com/python/typing/commit/60aa1e20c07af2f0165485a2cdbbd23a5265a194
- Released here: https://github.com/python/typing/commit/b53cb9a0a85576290fa549324d0673a1b3969897

I'll do a release.

inducer:
Thanks!

inducer:
https://pypi.org/project/pytools/2022.1.5/

PR #132, opened 2022-04-28
`Record.__eq__`: early exit for identical instances

kaushikcfd:


inducer:
Thanks! I guess I'll roll a release with this.

PR #133, opened 2022-05-04
UniqueNameGenerator.add_name(s?): new argument `existing_ok`

kaushikcfd:


pytools/__init__.py
```diff
@@ -2311,8 +2311,14 @@ def _name_added(self, name: str) -> None:
         """
         pass
 
-    def add_name(self, name: str) -> None:
-        if self.is_name_conflicting(name):
+    def add_name(self, name: str, *, existing_ok: bool = False) -> None:
```
inducer, line 2314:
```suggestion
    def add_name(self, name: str, *, conflicting_ok: bool = False) -> None:
```

inducer: The user can (as loopy does) define "conflicting' to mean different things. (In loopy, that's used to keep `_s[0-9]` of existing names suffixes clear.)

kaushikcfd: Yep, thanks. That's more inline with the logic there. Fixed via https://github.com/inducer/pytools/compare/306b08970069df6001dd852a6018d35f0de05945..3a3f878ae32416f8fe89e5419f5751ea6de850cd.

inducer:
Will do a release once this merges.

inducer:
https://pypi.org/project/pytools/2022.1.7/

PR #134, opened 2022-05-17
add IgnoredForEqualityTag

matthiasdiener:
Needed by:
- [ ] https://github.com/inducer/pytato/pull/284
- [x] https://github.com/inducer/meshmode/pull/329

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -324,7 +324,7 @@ class Taggable:
                          if isinstance(tag, tag_t)})
 
     @memoize_method
-    def tags_not_of_type(self, tag_t: Type[TagT]) -> FrozenSet[TagT]:
+    def tags_not_of_type(self, tag_t: Type[TagT]) -> FrozenSet[Tag]:
         """
         Returns *self*'s tags that are not of type *tag_t*.
         """
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -244,6 +244,7 @@ class Taggable:
     .. automethod:: tagged
     .. automethod:: without_tags
     .. automethod:: tags_of_type
+    .. automethod:: tags_not_of_type
 
     .. versionadded:: 2021.1
     """
@@ -322,6 +323,27 @@ class Taggable:
                          for tag in self.tags
                          if isinstance(tag, tag_t)})
 
+    @memoize_method
+    def tags_not_of_type(self, tag_t: Type[TagT]) -> FrozenSet[TagT]:
+        """
+        Returns *self*'s tags that are not of type *tag_t*.
+        """
+        return frozenset({tag
+                         for tag in self.tags
+                         if not isinstance(tag, tag_t)})
+
+    def __eq__(self, __o: object) -> bool:
+        new_self_tags = self.tags_not_of_type(IgnoredForEqualityTag)
+        if isinstance(__o, Taggable):
+            new_other_tags = __o.tags_not_of_type(IgnoredForEqualityTag)
+            return new_self_tags == new_other_tags
+        else:
+            return super().__eq__(__o)
+
+    def __hash__(self) -> int:
+        return hash(self.tags_not_of_type(IgnoredForEqualityTag))
+
+
 # }}}
 
 
@@ -330,15 +352,7 @@ class Taggable:
 class IgnoredForEqualityTag(Tag):
     """A type of tag whose value is purely informational and should not be used
     for equality comparison."""
-
-    def __eq__(self, __o: object) -> bool:
-        if isinstance(__o, IgnoredForEqualityTag):
-            return True
-        else:
-            return super().__eq__(__o)
-
-    def __hash__(self) -> int:
-        return hash(IgnoredForEqualityTag)
+    pass
 
 # }}}
 
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -354,7 +354,7 @@ class IgnoredForEqualityTag(Tag):
     :class:`Taggable`.
 
     When testing equality of two instances of :class:`Taggable`, the equality
-    of the :attr:`tags` of both instances is tested after removing all
+    of the ``tags`` of both instances is tested after removing all
     instances of :class:`IgnoredForEqualityTag`. Instances of
     :class:`IgnoredForEqualityTag` are removed for hashing instances of
     :class:`Taggable`.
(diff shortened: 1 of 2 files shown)


pytools/tag.py
```diff
@@ -324,6 +325,15 @@ def tags_of_type(self, tag_t: Type[TagT]) -> FrozenSet[TagT]:
 # }}}
 
 
+# {{{ IgnoredForEqualityTag
+
+class IgnoredForEqualityTag(Tag):
```
inducer, line 330: Please override `Taggable.__eq__` and include a test.

matthiasdiener: I'm not 100% sure what you had in mind, but how does [f921af0](https://github.com/inducer/pytools/pull/134/commits/f921af0e5be2055691132ea030d327b5e99caebe) look?

kaushikcfd: `__hash__` must also be overridden, I think.

matthiasdiener: What do you think of [4320ef3](https://github.com/inducer/pytools/pull/134/commits/4320ef352623b2cbe16917c5c6033a4dc5925f92) ?

kaushikcfd: That looks good, thanks!

inducer: That's not quite what I had in mind (I figured you'd override `Taggable.__eq__`), but it's an interesting approach. Unfortunately, I think there are two aspects that make it impractical:

- We now can't have more than one of these in the set of tags (they all compare equal) after all.
- They'll still prompt inequality between objects that do and do not have them.

kaushikcfd: The problem with the Taggable approach would be every downstream user would have to call `super().__eq__`, I think.

matthiasdiener: I don't see how we can resolve these issues here. Should we just use the localized approach in pytato (https://github.com/inducer/pytato/pull/284/commits/0794bdb4fd986549bce4ed6f463e31f34f995dec) @kaushikcfd @inducer ?

inducer: > The problem with the Taggable approach would be every downstream user would have to call `Taggable.__eq__`, I think.

Yes, IMO that's expected, and a reasonable thing to do anyhow.

inducer: FWIW, we would need to investigate how dataclasses handle this.

inducer: As you can tell, I'm not a fan of the localized approach. We can either do this once, and consistently, or we'll keep hackering around with this nonsense forever.

pytools/tag.py
```diff
@@ -321,6 +323,37 @@ def tags_of_type(self, tag_t: Type[TagT]) -> FrozenSet[TagT]:
...
+
+
+# }}}
+
+
+# {{{ IgnoredForEqualityTag
+
+class IgnoredForEqualityTag(Tag):
+    """A type of tag whose value is purely informational and should not be used
+    for equality comparison."""
```
inducer, line 354: Be more precise here: What does "not be used for equality comparison" mean? Use words like presence, absence, subclass, equality of the actual tags.

test/test_pytools.py
```diff
@@ -664,6 +664,35 @@ def test_unique_name_gen_conflicting_ok():
     ung.add_names({"a", "b", "c"}, conflicting_ok=True)
 
 
+def test_ignoredforequalitytag():
+    from pytools.tag import IgnoredForEqualityTag, Tag, Taggable
+
+    # Need a subclass that defines the copy function in order to test.
+    class TaggableWithCopy(Taggable):
+
+        def copy(self, **kwargs):
```
inducer, line 673: Don't you want a `_with_new_tags` here?

test/test_pytools.py
```diff
@@ -664,6 +664,35 @@ def test_unique_name_gen_conflicting_ok():
...
+
+    class Eq2(IgnoredForEqualityTag):
+        pass
+
+    class Eq3(Tag):
+        pass
+
+    eq1 = TaggableWithCopy(frozenset([Eq1()]))
+    eq2 = TaggableWithCopy(frozenset([Eq2()]))
+    eq3 = TaggableWithCopy(frozenset([Eq3()]))
```
inducer, line 687: Also test presence and absence.

pytools/tag.py
matthiasdiener: What do you think of [15e9ce6](https://github.com/inducer/pytools/pull/134/commits/15e9ce6fd18d0e0aaad2ffa7db9ba9b9d238b0ba)?

test/test_pytools.py
matthiasdiener: Done in [9866d7d](https://github.com/inducer/pytools/pull/134/commits/9866d7d10ad7ccab057f2aa78130a8b31a06be23)

test/test_pytools.py
matthiasdiener: Done in [9866d7d](https://github.com/inducer/pytools/pull/134/commits/9866d7d10ad7ccab057f2aa78130a8b31a06be23)

pytools/tag.py
```diff
@@ -321,6 +323,44 @@ def tags_of_type(self, tag_t: Type[TagT]) -> FrozenSet[TagT]:
...
+    @memoize_method
+    def tags_not_of_type(self, tag_t: Type[TagT]) -> FrozenSet[Tag]:
+        """
+        Returns *self*'s tags that are not of type *tag_t*.
+        """
+        return frozenset({tag
+                         for tag in self.tags
+                         if not isinstance(tag, tag_t)})
+
+    def __eq__(self, __o: object) -> bool:
```
inducer, line 335: Double-underscore identifiers are reserved in Python, AFAIK.

matthiasdiener: Fixed in [b781d63](https://github.com/inducer/pytools/pull/134/commits/b781d6302ca0294fb1534fe2177fa5eef95111ce).\

Interestingly, `def __eq__(self, __o: object) -> bool` is the default signature of `__eq__` that my editor (VS Code) suggests.

inducer:
Thanks, LGTM!

inducer:
I'll do a release once this merges.

matthiasdiener:
Thank you!

inducer:
https://pypi.org/project/pytools/2022.1.8/

PR #135, opened 2022-05-19
add WeakUniqueTag

matthiasdiener:
WeakUniqueTag - A tag that can be only set once, subsequent attempts to add a tag instance of the same subclass will be silently ignored (in contrast to UniqueTag, which raises an exception)

This might make code such as https://github.com/inducer/meshmode/pull/329 cleaner.


inducer:
I'm sort of never a fan of silently ignoring anything.

matthiasdiener:
> I'm sort of never a fan of silently ignoring anything.

Ok, I see. With the new version of https://github.com/inducer/meshmode/pull/329, this becomes less important. Closing.

PR #145, opened 2022-06-18
Compute topological sort with dynamic key

### LINKED ISSUE #143: [Feature-request] Topological sort with keying function that exposes internal state
Consider the DAG below:
```python
from pytools.graph import compute_topological_order

dag = {"A": {"C"},
       "B": {"E"},
       "C": {"D"},
       "D": set(),
       "E": set(),
       }

colors = {"A": "red",
          "B": "red",
          "C": "blue",
          "D": "red",
          "E": "blue"
          }

sorted_nodes = compute_topological_order(dag, key=lambda x: x)
print(sorted_nodes)                            # ['A', 'B', 'C', 'D', 'E']
print([colors[node] for node in sorted_nodes]) # ['red', 'red', 'blue', 'red', 'blue']
```

My application needs a valid topological sort that minimizes the jumps from 'red'<->'blue' nodes in the final schedule. And so for this problem, the topological sort order I'm after would be `['A', 'B', 'C', 'E', 'D']`. I don't think that's currently possible with the current interface of `compute_topological_order`.

kaushikcfd:
Closes #143.


pytools/graph.py
```diff
@@ -199,26 +203,51 @@ class HeapEntry:
     Only needs to define :func:`pytools.graph.__lt__` according to
     <https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Lib/heapq.py#L135-L138>.
     """
-    def __init__(self, node, key):
-        self.node = node
-        self.key = key
+    node: Any
+    key: Any
```
inducer, line 207: Maybe introduce type variables for key and node, for increased expressiveness of the type annotation?

pytools/graph.py
```diff
@@ -199,26 +203,51 @@ class HeapEntry:
...
-        extract a comparison key from each node of the *graph*.
+    :arg trigger_key_update: A function called after scheduling a node in
+        *graph* that takes in an instance of :class:`TopologicalOrderState`
+        corresponding to the scheduling state at that point and returns whether
+        the comparison keys corresponding to the nodes be updated.
+
+    :arg get_key: A callable  called when *trigger_key_update*
+        returns *True*. Takes in an instance of :class:`TopologicalOrderState`
+        and returns another callable that accepts node as an argument and returns the
+        comparison key corresponding to the node.
```
inducer, line 250: Couldn't this be a single callable that either performs the update or says "I've not done anything"?

inducer (commented):
Thanks for working on this! While I agree that the operation introduced here is mostly well-defined, the approach seems a bit "brutal". :slightly_smiling_face: The potential $O(n^2)$ cost also points in the same direction. As a result, I'm hesitatnt to cement this into an interface that we might need to support longer term. Maybe just let this live in loopy (where I imagine you're using this) until a more "final" approach emerges.

PR #147, opened 2022-06-26
pdict: When building dataclass hash, include type

inducer:


matthiasdiener:
Out of curiosity, why is this including the type of the `key_hash`, and not the `key`?

inducer:
That's a bug.

matthiasdiener:
I'll prepare a patch.

matthiasdiener:
Done in #196.

PR #150, opened 2022-10-16
add strtobool()

matthiasdiener:
Based on the now-deprecated [distutils.util.strtobool](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool).

Intended to make code like https://github.com/inducer/pytato/blob/a0fa84a5ca17a32f57431b8d67d0383f4c94839c/pytato/__init__.py#L32-L40 more robust.

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -2917,18 +2917,24 @@ def sphere_sample_fibonacci(
 def strtobool(val: Optional[str], default: Optional[bool] = None) -> bool:
     """Convert a string representation of truth to True or False.
     True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
-    are 'n', 'no', 'f', 'false', 'off', and '0'.  If *default* is None,
-    raises ValueError if *val* is anything else. If *val* is None and
-    *default* is not None, returns *default*. Based on
-    :func:`distutils.util.strtobool`.
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Uppercase versions are
+    also accepted. If *default* is None, raises ValueError if *val* is anything
+    else. If *val* is None and *default* is not None, returns *default*.
+    Based on :func:`distutils.util.strtobool`.
 
     :param val: Value to convert.
     :param default: Value to return if *val* is None.
 
-    :returns: True or False.
+    :returns: Truth value of *val*.
     """
     if val is None and default is not None:
         return default
+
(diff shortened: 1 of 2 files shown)


alexfikl:
Wouldn't it be easier to handle that the way `pyopencl` does it
https://github.com/inducer/pyopencl/blob/497cd1562454e6c2880d969e88b932cfc3b93648/pyopencl/__init__.py#L362
? i.e. its true if the environment variable is defined and false otherwise.

matthiasdiener:
> Wouldn't it be easier to handle that the way `pyopencl` does it
> 
> https://github.com/inducer/pyopencl/blob/497cd1562454e6c2880d969e88b932cfc3b93648/pyopencl/__init__.py#L362
> 
> ? i.e. its true if the environment variable is defined and false otherwise.

We could do that, but imho the interface is confusing. People have been using PYOPENCL_NO_CACHE=0 to "enable" the cache...

pytools/__init__.py
```diff
@@ -2907,6 +2912,34 @@ def sphere_sample_fibonacci(
...
+    :returns: True or False.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        if default is None:
+            raise ValueError(f"invalid truth value {val}")
```
inducer, line 2936: Add quotes.

pytools/__init__.py
```diff
@@ -2907,6 +2912,34 @@ def sphere_sample_fibonacci(
 # }}}
 
 
+# {{{ strtobool
+
+def strtobool(val: str, default: Optional[bool] = None) -> bool:
```
inducer, line 2917: I could see returning default if `val` is `None`, but returning it if the value is not understood is unintuitive and potentially dangerous IMO.

matthiasdiener: Done in 83d121e

pytools/__init__.py
```diff
@@ -2907,6 +2912,38 @@ def sphere_sample_fibonacci(
...
```
alexfikl, line 2917:
```suggestion
def strtobool(val: Optional[str], default: Optional[bool] = None) -> bool:
```

inducer:
Thanks!

PR #151, opened 2022-10-20
add get_graph_dot_code and show_dot

matthiasdiener:


setup.py
```diff
@@ -40,6 +40,7 @@
           "platformdirs>=2.2.0",
           "dataclasses>=0.7;python_version<='3.6'",
           "typing_extensions>=4.0; python_version<'3.11'",
+          "graphviz",
```
inducer, line 43: Hard dependency on graphviz: not happening. Why not move/use the existing functionality for visualizing dot output, e.g. in pymbolic/pytato?

matthiasdiener: What do you think of [cab9e95](https://github.com/inducer/pytools/pull/151/commits/cab9e952cae5b10254679aa79097c9d6a7ca010d)? This uses the dot visualization from pymbolic.

pytools/graph.py
```diff
@@ -369,4 +370,40 @@ def compute_induced_subgraph(graph: Mapping[T, Set[T]],
...
+
+    res = "digraph mygraph {\n"
+
+    for node, targets in graph.items():
+        array_to_id[node] = id_gen()
+        res += f'{array_to_id[node]} [label="{dot_escape(repr(node))}"];\n'
+        for t in targets:
+            array_to_id[t] = id_gen()
+            res += f'{array_to_id[t]} [label="{dot_escape(repr(t))}"];\n'
+            res += f"{array_to_id[node]} -> {array_to_id[t]};\n"
```
inducer, line 400: This has quadratic complexity because of the inefficient string assembly. Instead, build a list and `join` it.

pytools/graph.py
```diff
@@ -369,4 +370,40 @@ def compute_induced_subgraph(graph: Mapping[T, Set[T]],
...
+        that are connected to the node by outgoing edges.
+    """
+    from pytools import UniqueNameGenerator
+    id_gen = UniqueNameGenerator(forced_prefix="mynode")
+
+    from pytools.dot import dot_escape, show_dot
+
+    array_to_id = {}
+
+    res = "digraph mygraph {\n"
```
inducer, line 392: Use f-strings better here, maybe

```python
result = "digraph mygraph {{\n{ content }\n}}"
```
So that open/close brace remain close in code.

setup.py
inducer: Could you prepare a PR to pymbolic that lets it make use of this?

pytools/graph.py
matthiasdiener: Is 4dd3b81 what you had in mind?

pytools/graph.py
matthiasdiener: Should be fixed in 
4dd3b81

pytools/graph.py
```diff
@@ -369,4 +370,54 @@ def compute_induced_subgraph(graph: Mapping[T, Set[T]],
...
+         for node in node_to_id.keys()])
+
+    content += "\n"
+
+    content += "\n".join(
+        [f"{node_to_id[node]} -> {node_to_id[t]};" for (node, t) in edges])
+
+    res = f"digraph mygraph {{\n{ content }\n}}\n"
+
+    show_dot(res, output_to)
```
alexfikl, line 417: Calling an external executable here doesn't seem necessary, since
```
show_dot(visualize_graph(...), output_to="svg")
```
is pretty short. Any specific reason to bundle these together?

matthiasdiener: My main motivation was based on my impression that few users actually want to "get" the dot code, but rather want to visualize it. But you're right, it isn't too much extra code, and makes the test simpler, so I'll change it to `get_graph_dot_code` and remove the call to `show_dot`.

alexfikl: That's fair enough. I'm not a user of most of this stuff, so take my comment with a sizeable grain of salt!

(The use case I had in mind was just writing the dot code to a file for later)

setup.py
matthiasdiener: Done in https://github.com/inducer/pymbolic/pull/115

pytools/graph.py
matthiasdiener: Further improved in https://github.com/inducer/pytools/pull/151/commits/48169a565b96dc80a4e5d80453f3aa5e942e27ca

pytools/graph.py
matthiasdiener: No worries, I think the API is much cleaner now! Thanks for the suggestion!

matthiasdiener:
This is ready for another review

pytools/dot.py
```diff
@@ -0,0 +1,130 @@
+__copyright__ = """
```
inducer, line 1: Dumb question: should this be `pytools.graph.graphviz`?

matthiasdiener: Dunno - the function names usually refer to dot: `show_dot`, `get_dot_code`, ...

pytools/graph.py
```diff
@@ -369,4 +370,45 @@ def compute_induced_subgraph(graph: Mapping[T, Set[T]],
 
 # }}}
 
+
+# {{{
+
+def get_graph_dot_code(graph: Mapping[T, Iterable[T]]) -> str:
```
inducer, line 376: That type annotation does not make sense. `Iterables` are only guaranteed to allow iteration once. Did you mean `Sequence`?

matthiasdiener: Sorry, I'm not sure I understand - is there a place in the function that requires multiple iteration on the values in `graph`? Also,  `Sequence` would disallow `set`s I think, so should it be a `Collection` instead?

inducer: > Sorry, I'm not sure I understand - is there a place in the function that requires multiple iterations over the values in `graph` (I can only see the `for t in targets` line)?

I mean, if so, you'd be left with a dictionary full of exhausted iterators (e.g.).

> so should it be a `Collection` instead?

Yes, that's better. Thanks!

matthiasdiener: Oh, I see! Should we change the signature of the other functions in this file that use Iterable to Collection also?

inducer: Yes, I think that'd be better.

matthiasdiener: Done in 93c12a5

matthiasdiener:
This is ready for review.

pytools/dot.py
inducer: While `dot` is certainly the most-used of the graphviz tools, there are way more tools in the package that this could easily support.

The fact that we've made a certain mistake in the past doesn't mean that we have to keep making it. :slightly_smiling_face:

pytools/graph.py
```diff
@@ -397,6 +398,50 @@ def compute_induced_subgraph(graph: Mapping[T, Set[T]],
 # }}}
 
 
+# {{{ get_graph_dot_code
+
+def get_graph_dot_code(graph: Mapping[T, Collection[T]]) -> str:
```
inducer, line 403: Maybe support labeling the edges, to support the functionality we discussed Friday.

pytools/dot.py
```diff
@@ -0,0 +1,130 @@
...
+            assert output_to == "browser"
+
+            from webbrowser import open as browser_open
+            browser_open("file://" + full_svg_file_name)
+    else:
+        raise ValueError("`output_to` can be one of 'xwindow', 'browser', or 'svg',"
+                         f" got '{output_to}'")
+
+    return None
+# }}}
```
inducer, line 130: If you have these, also have a
```
# vim: foldmethod=marker
```

matthiasdiener: added in [4ec6ead](https://github.com/inducer/pytools/pull/151/commits/4ec6eadd0b2768678e2646403f61fb317a00acc8)

pytools/graph.py
matthiasdiener: Added in [258d7e8](https://github.com/inducer/pytools/pull/151/commits/258d7e87db0462f15e9198242fbb236113343bc5)

pytools/dot.py
matthiasdiener: I renamed the file to `pytools.graphviz` in [6f178ce](https://github.com/inducer/pytools/pull/151/commits/6f178cede6273cd8f33f7121a11a9efc1a15e8a3). Since `show_dot` etc. aren't really specific to `graph`, I think putting them under `pytools.graph` may not be the best fit.

pytools/graph.py
```diff
@@ -410,6 +411,62 @@ def compute_induced_subgraph(graph: Mapping[NodeT, Set[NodeT]],
 # }}}
 
 
+# {{{ get_graph_dot_code
+
+def get_graph_dot_code(graph: Mapping[T, Collection[T]],
```
inducer, line 416:
```suggestion
def as_graphviz_dot(graph: GraphT[NodeT],
    node_labels: Optional[Callable[[NodeT], str]] = None,  # default str?
    edge_labels: Optional[Callable[[NodeT, NodeT], str]] = None)
```

pytools/graph.py
```diff
@@ -410,6 +411,62 @@ def compute_induced_subgraph(graph: Mapping[NodeT, Set[NodeT]],
...
+    from pytools import UniqueNameGenerator
+    id_gen = UniqueNameGenerator(forced_prefix="mynode")
+
+    from pytools.graphviz import dot_escape
+
+    if edge_labels is None:
+        edge_labels = {}
+
+    node_to_id = {}
+    edges = []
```
inducer, line 442: I think this works without explicitly building `edges`.

matthiasdiener: Fixed in d8c6ac9

pytools/graph.py
matthiasdiener: done in [d84cb73](https://github.com/inducer/pytools/pull/151/commits/d84cb7397ec007d3b6913c25fe16e3fc1fea3682)

inducer:
LGTM, thanks!

inducer:
https://pypi.org/project/pytools/2022.1.14/

PR #152, opened 2022-10-31
Add `reverse_graph`

inducer:
cc @matthiasdiener 

- [x] Needs tests

### DIFF
--- a/pytools/graph.py
+++ b/pytools/graph.py
@@ -29,6 +29,7 @@ __doc__ = """
 Graph Algorithms
 =========================
 
+.. autofunction:: reverse_graph
 .. autofunction:: a_star
 .. autofunction:: compute_sccs
 .. autoclass:: CycleError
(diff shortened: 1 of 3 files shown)


pytools/graph.py
```diff
@@ -45,13 +46,37 @@
...
+    Reverses a graph.
+
+    :param graph: A :class:`dict` representation of a directed graph, mapping each
+        node to other nodes to which it is connected by edges.
+    :returns: A :class:`dict` representing *graph* with edges reversed.
+    """
+    result: Dict[T, Set[T]] = {}
+
+    for node_key, edges in graph.items():
+        # Make sure every node is in the result even if it has no users
```
kaushikcfd, line 69:
```suggestion
        # Make sure every node is in the result even if it has no successors
```

pytools/graph.py
```diff
@@ -45,13 +46,37 @@
...
 
 
+# {{{ reverse_graph
+
+def reverse_graph(graph: Mapping[T, Collection[T]]) -> Dict[T, FrozenSet[T]]:
+    """
+    Reverses a graph.
+
+    :param graph: A :class:`dict` representation of a directed graph, mapping each
+        node to other nodes to which it is connected by edges.
```
kaushikcfd, line 63: Doesn't explicitly talk about the direction of the edge.

pytools/graph.py
```diff
@@ -45,13 +46,37 @@
...
+    :returns: A :class:`dict` representing *graph* with edges reversed.
+    """
+    result: Dict[T, Set[T]] = {}
+
+    for node_key, edges in graph.items():
```
kaushikcfd, line 68:
```suggestion
    for node_key, successors in graph.items():
```
"edges" makes it sounds like the values are a tuple pair.

kaushikcfd (approved):
LGTM!
Left some minor readability improvement suggestions.

pytools/graph.py
inducer: I thought about that! I ended up deciding that there was no need to document this, as it's just the reverse of what it was before, no matter which way they were supposed to be pointing initially.

kaushikcfd: Fair, sg!

inducer:
Thanks all!

PR #155, opened 2022-11-10
add validate_graph

matthiasdiener:


pytools/graph.py
```diff
@@ -394,4 +395,28 @@ def compute_induced_subgraph(graph: Mapping[T, Set[T]],
...
+        that are connected to the node by outgoing edges.
+    """
+    seen_nodes: Set[T] = set()
+
+    for children in graph.values():
+        seen_nodes.update(children)
+
+    if not seen_nodes <= graph.keys():
+        raise ValueError(
+            f"invalid graph: {graph.keys()}, seen nodes: {seen_nodes}")
```
inducer, line 418: This should probably mention the offending key. Also, I think printing the entire data structure is not a good idea: this could be very large.

inducer (commented):
Thanks! Just one minor point to fix here.

pytools/graph.py
matthiasdiener: Good idea - what do you think of 1aa6181?

inducer: LGTM

PR #159, opened 2022-11-17
CycleError: Report all nodes

matthiasdiener:


pytools/graph.py
```diff
@@ -212,10 +212,10 @@ class CycleError(Exception):
     """
     Raised when a topological ordering cannot be computed due to a cycle.
 
-    :attr node: Node in a directed graph that is part of a cycle.
+    :attr nodes: List of nodes in a directed graph that are part of a cycle.
     """
-    def __init__(self, node: T) -> None:
-        self.node = node
+    def __init__(self, nodes: List[T]) -> None:
+        self.nodes = nodes
```
inducer, line 218: This breaks backward compatibility.

inducer, line 218: It might make sense to override the `repr` to only show at most 10 or so nodes, to avoid massive output.

pytools/graph.py
```diff
@@ -299,8 +299,8 @@ def compute_topological_order(graph: Mapping[T, Collection[T]],
 
     if len(order) != total_num_nodes:
         # any node which has a predecessor left is a part of a cycle
-        raise CycleError(next(iter(n for n, num_preds in
-            nodes_to_num_predecessors.items() if num_preds != 0)))
+        raise CycleError(list(n for n, num_preds in
+            nodes_to_num_predecessors.items() if num_preds != 0))
```
inducer, line 303: It would be nice to provide this in traversable (rather than arbitrary) order, which is doable at $O(n)$ cost.

inducer:

kaushikcfd:
Also worth referring to the suggestions in this PR: https://gitlab.tiker.net/inducer/pytools/-/merge_requests/37.

pytools/graph.py
matthiasdiener: I'm not sure how to do that - how would we provide this in case there are multiple cycles in the graph?

pytools/graph.py
inducer: I think a better approach would be to make `node` a `@property` that spews a `DeprecationWarning`.

pytools/graph.py
```diff
@@ -213,9 +213,17 @@ class CycleError(Exception):
     Raised when a topological ordering cannot be computed due to a cycle.
 
     :attr node: Node in a directed graph that is part of a cycle.
+
+    :attr nodes: List of nodes in a directed graph that are part of a cycle.
```
inducer, line 217: Maybe `path` is a better name. Also state that we claim that `node[i+1]` in `path`  is a successor of `node[i]` (and test that!).

pytools/graph.py
matthiasdiener: #167 has some work in this direction

pytools/graph.py
matthiasdiener: Implemented in [8823bca](https://github.com/inducer/pytools/pull/159/commits/8823bcad13949b3a14f88c0fb495c9bd494e1101)

pytools/graph.py
matthiasdiener: Changed it to `paths` and added comment regarding the successors in [47aa316](https://github.com/inducer/pytools/pull/159/commits/47aa3164f3a5e04c9f16afe3bf23fd9cde226275).
Testing this is done in #167 right?

PR #160, opened 2022-11-17
Set Ubuntu CI image to 20.04

matthiasdiener:
Newer images have no Python 3.6 support:
https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

See e.g. https://github.com/inducer/pytools/actions/runs/3492840943/jobs/5847030018 for a failing CI run which tries to use Ubuntu 22.04 with Python 3.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,8 @@ on:
 jobs:
     flake8:
         name: Flake8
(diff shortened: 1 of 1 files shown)


inducer:
See #161 instead.

matthiasdiener:
> See #161 instead.

Ok, even better =)

PR #163, opened 2022-12-06
graph: add is_connected

matthiasdiener:


pytools/graph.py
```diff
@@ -420,4 +421,45 @@ def validate_graph(graph: Mapping[T, Collection[T]]) -> None:
...
+        for child in undirected_graph[node]:
+            if child not in visited:
+                dfs(child)
+
+    dfs(next(iter(graph.keys())))
+
+    if visited == graph.keys():
+        return True
+    else:
+        return False
```
inducer, line 462:
```suggestion
    return visited == graph.keys()
```

pytools/graph.py
```diff
@@ -420,4 +421,45 @@ def validate_graph(graph: Mapping[T, Collection[T]]) -> None:
...
+    the edge direction.
+
+    :arg graph: A :class:`collections.abc.Mapping` representing a directed
+        graph. The dictionary contains one key representing each node in the
+        graph, and this key maps to a :class:`collections.abc.Collection` of nodes
+        that are connected to the node by outgoing edges.
+    """
+    validate_graph(graph)
+
+    if len(graph.keys()) == 0:
```
inducer, line 439:
```suggestion
    if not graph:
```

pytools/graph.py
```diff
@@ -420,4 +421,45 @@ def validate_graph(graph: Mapping[T, Collection[T]]) -> None:
...
+
+def is_connected(graph: Mapping[T, Collection[T]]) -> bool:
+    """
+    Returns whether all nodes in *graph* are connected, ignoring
+    the edge direction.
+
+    :arg graph: A :class:`collections.abc.Mapping` representing a directed
+        graph. The dictionary contains one key representing each node in the
+        graph, and this key maps to a :class:`collections.abc.Collection` of nodes
+        that are connected to the node by outgoing edges.
```
inducer, line 435: By this point, I think we might be repeating this definition over and over. It might be sensible to refer to a central location instead.

pytools/graph.py
```diff
@@ -420,4 +421,45 @@ def validate_graph(graph: Mapping[T, Collection[T]]) -> None:
...
+    """
+    validate_graph(graph)
```
inducer, line 437: I don't love this: it imposes cost on the user even when they know that their graph is valid.

pytools/graph.py
matthiasdiener: What do you think of #166?

pytools/graph.py
matthiasdiener: Good point, I removed this in 99fd92b

inducer:
Looks great, thanks!

PR #166, opened 2022-12-07
graph: create GraphT

matthiasdiener:
Before:

image

After:

image

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -38,4 +38,6 @@ nitpick_ignore_regex = [
 
 nitpicky = True
 
-autodoc_type_aliases = {"GraphT": "GraphT"}
+autodoc_type_aliases = {"GraphT": "GraphT",
+                        "NodeT": "NodeT",
+                        }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -38,6 +38,6 @@ nitpick_ignore_regex = [
 
 nitpicky = True
 
-autodoc_type_aliases = {"GraphT": "GraphT",
-                        "NodeT": "NodeT",
+autodoc_type_aliases = {"GraphT": "pytools.graph.GraphT",
+                        "NodeT": "pytools.graph.NodeT",
                         }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -37,3 +37,5 @@ nitpick_ignore_regex = [
         ]
 
 nitpicky = True
+
+autodoc_type_aliases = {"GraphT": "pytools.graph.GraphT"}
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -38,4 +38,4 @@ nitpick_ignore_regex = [
 
 nitpicky = True
 
-autodoc_type_aliases = {"GraphT": "pytools.graph.GraphT"}
+autodoc_type_aliases = {"GraphT": "GraphT"}
--- a/pytools/graph.py
+++ b/pytools/graph.py
@@ -62,9 +62,9 @@ Type Variables Used
     is included as a key in the graph.
 """
 
-from typing import (Collection, TypeVar, Mapping, List, Optional, Any,
+from typing import (Collection, Mapping, List, Optional, Any,
                     Callable, Set, MutableSet, Dict, Iterator, Tuple,
-                    Type, Hashable)
+                    Hashable)
 
 try:
     from typing import TypeAlias
@@ -72,7 +72,7 @@ except ImportError:
     from typing_extensions import TypeAlias
 
 
-NodeT = TypeVar("NodeT", bound=Hashable)
+NodeT: TypeAlias = Hashable
 
 
 GraphT: TypeAlias = Mapping[NodeT, Collection[NodeT]]
@@ -88,7 +88,7 @@ def reverse_graph(graph: GraphT) -> GraphT:
 
     :returns: A :class:`dict` representing *graph* with edges reversed.
     """
-    result: Dict[Type[GraphT], Set[Type[GraphT]]] = {}
+    result: Dict[NodeT, Set[NodeT]] = {}
 
     for node_key, successor_nodes in graph.items():
         # Make sure every node is in the result even if it has no successors
@@ -424,7 +424,7 @@ def validate_graph(graph: GraphT) -> None:
 
     :arg graph: An instance of :class:`GraphT`.
     """
-    seen_nodes: Set[Type[GraphT]] = set()
+    seen_nodes: Set[NodeT] = set()
 
     for children in graph.values():
         seen_nodes.update(children)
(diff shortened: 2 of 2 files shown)


inducer (approved):
This looks great. Thanks!

matthiasdiener:
> This looks great. Thanks!

Thanks! I did a few minor fixes and changes, please let me know what you think.

pytools/graph.py
```diff
@@ -43,29 +49,46 @@
...
+                    Callable, Set, MutableSet, Dict, Iterator, Tuple, FrozenSet,
+                    Type)
+
+try:
+    from typing import TypeAlias
+except ImportError:
+    from typing_extensions import TypeAlias
+
+
+NodeT = TypeVar("NodeT")
```
alexfikl, line 75:
```suggestion
NodeT = TypeVar("NodeT", bound=Hashable)
```
? Not sure if that brings anything, but should be clearer since it's used as a key in the mapping.

pytools/graph.py
```diff
@@ -43,29 +49,46 @@
...
 
 # {{{ reverse_graph
 
-def reverse_graph(graph: Mapping[T, Collection[T]]) -> Dict[T, FrozenSet[T]]:
+def reverse_graph(graph: GraphT) -> Dict[NodeT, FrozenSet[NodeT]]:
     """
-    Reverses a graph.
+    Reverses a graph *graph*.
+
+    :arg graph: An instance of :class:`GraphT`.
```
alexfikl, line 87:
```suggestion
```
? The arg doc doesn't bring much now that the type annotation links to the `GraphT` docs.

pytools/graph.py
```diff
@@ -43,29 +49,46 @@
...
 
 
-T = TypeVar("T")
+GraphT: TypeAlias = Mapping[NodeT, Collection[NodeT]]
 
 
 # {{{ reverse_graph
 
-def reverse_graph(graph: Mapping[T, Collection[T]]) -> Dict[T, FrozenSet[T]]:
+def reverse_graph(graph: GraphT) -> Dict[NodeT, FrozenSet[NodeT]]:
```
alexfikl, line 83:
```suggestion
def reverse_graph(graph: GraphT) -> GraphT:
```
? Does that work or not because of the `Collection` vs `FrozenSet`?

pytools/graph.py
```diff
@@ -310,13 +331,13 @@ def compute_topological_order(graph: Mapping[T, Collection[T]],
 
 # {{{ compute transitive closure
 
-def compute_transitive_closure(graph: Mapping[T, MutableSet[T]]) -> (
-        Mapping[T, MutableSet[T]]):
+def compute_transitive_closure(graph: Mapping[NodeT, MutableSet[NodeT]]) -> (
```
alexfikl, line 334: This doesn't work with `GraphT`? :cry:

alexfikl (commented):
Gave this a quick glance out of curiosity and left some comments (nitpicks, so not meant to keep you from merging at all!)

pytools/graph.py
matthiasdiener: mypy doesn't like this: `pytools/graph.py:361: error: "Collection[Any]" has no attribute "add"  [attr-defined]` :-(

pytools/graph.py
matthiasdiener: Good point, I added this in 4c8ad32

pytools/graph.py
matthiasdiener: I agree. My thinking was that each argument in the function definition should also be listed here, since otherwise it might look like the two argument lists don't match. I don't have a strong preference either way though.

pytools/graph.py
matthiasdiener: It does work actually, I changed this in [dccdf19](https://github.com/inducer/pytools/pull/166/commits/dccdf19bdf3a6775235fe5aab5d381368e027bd6)

pytools/graph.py
```diff
@@ -80,15 +80,15 @@
...
+def reverse_graph(graph: GraphT) -> GraphT:
     """
     Reverses a graph *graph*.
 
     :arg graph: An instance of :class:`GraphT`.
 
     :returns: A :class:`dict` representing *graph* with edges reversed.
     """
-    result: Dict[NodeT, Set[NodeT]] = {}
+    result: Dict[Type[GraphT], Set[Type[GraphT]]] = {}
```
matthiasdiener, line 91: Without this change, my complains:

```
pytools/graph.py:91: error: Type variable "pytools.graph.NodeT" is unbound  [valid-type]
pytools/graph.py:91: note: (Hint: Use "Generic[NodeT]" or "Protocol[NodeT]" base class to bind "NodeT" inside a class)
pytools/graph.py:91: note: (Hint: Use "NodeT" in function signature to bind "NodeT" inside a function)
```

alexfikl: That looks very weird.. `Type[GraphT]` should just be `dict` (i.e. the class).

I would say it's not worth complicating these if mypy isn't helpful..

alexfikl (approved):

alexfikl:
By the way, feel free to resolve my comments (I don't have authority)!

pytools/graph.py
matthiasdiener: With 530a7dc, this isn't necessary anymore.

inducer (commented):
Thanks! Just two more small nits, then this is ready to go.

pytools/graph.py
matthiasdiener: Ok, I removed them in a564ef2

doc/conf.py
```diff
@@ -37,3 +37,7 @@
         ]
 
 nitpicky = True
+
+autodoc_type_aliases = {"GraphT": "GraphT",
+                        "NodeT": "NodeT",
```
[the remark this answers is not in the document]
matthiasdiener: It seems to work fine either way; I felt that having the full name is a bit too verbose here, but I don't have a strong opinion either way.

Here is how it looks like with the full name:

image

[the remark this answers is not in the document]
matthiasdiener: Changed to full name in 9e29f53

inducer:
Thanks!

pytools/graph.py
```diff
@@ -43,29 +49,44 @@
...
+                    Hashable)
 
+try:
+    from typing import TypeAlias
+except ImportError:
+    from typing_extensions import TypeAlias
 
-T = TypeVar("T")
+
+NodeT: TypeAlias = Hashable
```
inducer, line 75: So I missed that the node type went from a type variable to just `Hashable`, which is essential, and without which type checking in a bunch of downstream packages, e.g. here:

https://gitlab.tiker.net/inducer/pytato/-/jobs/498112

The correct thing would have been for `NodeT` to remain a type variable, and for `GraphT` to become a parametric type alias. I will push a fix in a bit.

PR #167, opened 2022-12-08
Better cycle detection

### LINKED ISSUE #165: graph: Misleading reporting of node in cycle for CycleError
Initially reported by @majosm.

With this code:

```python
graph={1: set(), 5: { 1, 8}, 8: {5}, }

from pytools.graph import compute_topological_order,
compute_topological_order(graph)
```

image

`compute_topological_order` reports:

```
  File "/Users/mdiener/Work/emirge/pytools/pytools/graph.py", line 303, in compute_topological_order
    raise CycleError(next(iter(n for n, num_preds in
pytools.graph.CycleError: 1
```

matthiasdiener:
Fixes #165

### DIFF
--- a/pytools/graph.py
+++ b/pytools/graph.py
@@ -299,9 +299,26 @@ def compute_topological_order(graph: Mapping[T, Collection[T]],
                 heappush(heap, HeapEntry(child, keyfunc(child)))
 
     if len(order) != total_num_nodes:
-        # any node which has a predecessor left is a part of a cycle
-        raise CycleError(next(iter(n for n, num_preds in
-            nodes_to_num_predecessors.items() if num_preds != 0)))
+        # There is a cycle in the graph
+        try:
+            validate_graph(graph)
+        except ValueError:
+            # Graph is invalid, we can't compute SCCs or return a meaningful node
+            # that is part of a cycle
+            raise CycleError(None)
+
+        sccs = compute_sccs(graph)
+        cycles = [scc for scc in sccs if len(scc) > 1]
+
+        if cycles:
+            # Cycles that are not self-loops
+            node = cycles[0][0]
+        else:
+            # Self-loop SCCs also have a length of 1
+            node = next(iter(n for n, num_preds in
+                             nodes_to_num_predecessors.items() if num_preds != 0))
+
+        raise CycleError(node)
 
     return order
 
--- a/test/test_graph_tools.py
+++ b/test/test_graph_tools.py
@@ -395,6 +395,25 @@ def test_is_connected():
     assert is_connected({})
 
 
+def test_cycle_detection():
+    from pytools.graph import compute_topological_order, CycleError
+
+    # Non-Self Loop
+    graph = {1: {}, 5: {1, 8}, 8: {5}}
+    with pytest.raises(CycleError, match="5"):
+        compute_topological_order(graph)
+
+    # Self-Loop
+    graph = {1: {1}, 5: {8}, 8: {}}
+    with pytest.raises(CycleError, match="1"):
+        compute_topological_order(graph)
+
+    # Invalid graph with loop
+    graph = {1: {42}, 5: {8}, 8: {5}}
+    with pytest.raises(CycleError, match="None"):
+        compute_topological_order(graph)
+
+
 if __name__ == "__main__":
     if len(sys.argv) > 1:
         exec(sys.argv[1])


matthiasdiener:
This is ready for a first look.

pytools/graph.py
```diff
@@ -299,9 +299,26 @@ def compute_topological_order(graph: Mapping[T, Collection[T]],
...
+
+        if cycles:
+            # Cycles that are not self-loops
+            node = cycles[0][0]
+        else:
+            # Self-loop SCCs also have a length of 1
+            node = next(iter(n for n, num_preds in
+                             nodes_to_num_predecessors.items() if num_preds != 0))
+
+        raise CycleError(node)
```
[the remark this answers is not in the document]
matthiasdiener: I added a new API, `find_cycles()` in [27255a5](https://github.com/inducer/pytools/pull/167/commits/27255a589fc4d77deabd3463a81f52d603cffb22) and use it here. What do you think?

pytools/graph.py
```diff
@@ -240,6 +241,42 @@ def __init__(self, node: NodeT) -> None:
...
+def find_cycles(graph: GraphT) -> List[List[NodeT]]:
+    """
+    Find all cycles in *graph* using DFS.
+
+    :returns: A :class:`list` in which each element represents another :class:`list`
+        of nodes that form a cycle.
+    """
+    def dfs(node: NodeT, path: List[NodeT]) -> List[NodeT]:
+        # Cycle detected
+        if visited[node] == 1:
```
inducer, line 253: Make 0, 1, 2 an `enum`?

pytools/graph.py
```diff
@@ -240,6 +241,42 @@ def __init__(self, node: NodeT) -> None:
...
 
 
+def find_cycles(graph: GraphT) -> List[List[NodeT]]:
+    """
+    Find all cycles in *graph* using DFS.
+
+    :returns: A :class:`list` in which each element represents another :class:`list`
+        of nodes that form a cycle.
+    """
+    def dfs(node: NodeT, path: List[NodeT]) -> List[NodeT]:
```
inducer, line 251: Constructing `path` just in case is wasteful IMO: The path could be collected as you return, if a cycle is found.

pytools/graph.py
```diff
@@ -240,6 +241,42 @@ def __init__(self, node: NodeT) -> None:
...
+
+    visited = {node: 0 for node in graph.keys()}
+
+    res = []
+
+    for node in graph:
+        if not visited[node]:
+            cycle = dfs(node, [])
+            if cycle:
+                res.append(cycle)
```
inducer, line 275: Finding all cycles is probably too much... all the use cases I can think of will only want one. I could see this as an option, maybe.

pytools/graph.py
```diff
@@ -240,6 +241,42 @@ def __init__(self, node: NodeT) -> None:
         self.node = node
 
 
+def find_cycles(graph: GraphT) -> List[List[NodeT]]:
```
inducer, line 244: Could use this in `contains_cycle`.

pytools/graph.py
```diff
@@ -257,14 +294,16 @@ def __lt__(self, other: "HeapEntry") -> bool:
...
-                              -> List[NodeT]:
+                              key: Optional[Callable[[NodeT], Any]] = None,
+                              verbose_cycle: bool = True) -> List[NodeT]:
     """Compute a topological order of nodes in a directed graph.
 
     :arg key: A custom key function may be supplied to determine the order in
         break-even cases. Expects a function of one argument that is used to
         extract a comparison key from each node of the *graph*.
 
+    :arg verbose_cycle: Verbose reporting in case *graph* contains a cycle.
```
inducer, line 305: Say what this means.

pytools/graph.py
```diff
@@ -316,9 +355,17 @@ def compute_topological_order(graph: GraphT,
...
+        if not verbose_cycle:
+            raise CycleError(None)
+
+        try:
+            cycles = find_cycles(graph)
+        except KeyError:
+            # Graph is invalid
+            raise CycleError(None)
+        else:
+            raise CycleError(cycles[0][0])
```
inducer, line 368: Add to the documentation of `CycleError` what the value might mean.

pytools/graph.py
matthiasdiener: Done in 902e9ab

pytools/graph.py
matthiasdiener: I made it an optional parameter in 2ff7def. My thinking was that for the use case I had in mind, checking the validity of our DAG in pytato, having all cycles instead of just one may be beneficial.

pytools/graph.py
matthiasdiener: Hmm, the current doc of `CycleError` has `:attr node: Node in a directed graph that is part of a cycle.` - I'm not sure what else to add there.

pytools/graph.py
matthiasdiener: Done in 18d241b

pytools/graph.py
matthiasdiener: What do you think of a95c4c0?

pytools/graph.py
matthiasdiener: What do you think of [96bd3a6](https://github.com/inducer/pytools/pull/167/commits/96bd3a66e13aca8adcdb8cdee63965dd2bba15f9)?

pytools/graph.py
```diff
@@ -240,6 +243,52 @@ def __init__(self, node: NodeT) -> None:
         self.node = node
 
 
+class NodeState(Enum):
+    WHITE = 0  # Not visited yet
+    GREY = 1   # Currently visiting
+    BLACK = 2  # Done visiting
```
inducer, line 249: Why not use descriptive names for the node state?

pytools/graph.py
```diff
@@ -240,6 +243,52 @@ def __init__(self, node: NodeT) -> None:
         self.node = node
 
 
+class NodeState(Enum):
```
inducer, line 246:
```suggestion
class _NodeState(Enum):
```

inducer:
It seems that only one of #167 (this) and #159 should get merged. Which one would you like this to be?

pytools/graph.py
matthiasdiener: If you prefer, I can rename these, but I thought white/grey/black were standard labels in DFS (see e.g. http://www.cs.cmu.edu/afs/cs/academic/class/15750-s17/ScribeNotes/lecture9.pdf)

matthiasdiener:
> It seems that only one of #167 (this) and #159 should get merged. Which one would you like this to be?

I think these PRs are orthogonal to each other. This PR adds a new cycle detection function and uses it to correct the cycle reporting in `compute_topological_order` (it still only reports one node in a cycle there). #159 modifies the CycleError to contain all nodes part of cycles. Imho, #167 should be merged before #159.

pytools/graph.py
matthiasdiener: Thanks, done in 85a4a29.

matthiasdiener:
This does not work correctly in all cases, so I've set it back to draft now-

PR #171, opened 2023-01-14
Fix bugbear B028 errors

alexfikl:
They added a new check in the latest version
https://github.com/PyCQA/flake8-bugbear/issues/319

This fixes all the occurrences in `pytools`, but I'm not sure if it's better to disable it in other places. We used this fairly extensively :cry:

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -255,18 +255,19 @@ def deprecate_keyword(oldkey: str,
         def inner_wrapper(*args, **kwargs):
             if oldkey in kwargs:
                 if newkey is None:
-                    warn(f"The '{oldkey}' keyword is deprecated and will "
+                    warn(f"The {oldkey!r} keyword is deprecated and will "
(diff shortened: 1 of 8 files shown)


inducer:
Thanks for doing this. That said, I'm altogether not sure about this, so I'm inclined to hold off/disable. One example that gives me pause is pymbolic expression. `f"'{expr}'" uses the `str()` syntax with quotes, which is nice, whereas the `repr()` is unusable for all but the smallest examples. If nothing else, I'm not sure enough that this is a win to want to spend time "fixing" this everywhere. I might disable this in ci-support for now.

alexfikl:
Yeah, agreed. This was also awkward in the graphivz printing code. It seems like there are plenty of legitimate uses..

alexfikl:
This check seems to be on its way out, so I'll just close this!
https://github.com/PyCQA/flake8-bugbear/issues/331

PR #172, opened 2023-02-27
refdebug: concatenate tuple with tuple

matthiasdiener:
Was causing the error `TypeError: can only concatenate tuple (not "list") to tuple`


inducer:
Thx!

PR #174, opened 2023-03-08
typedump: type improvements

matthiasdiener:
- show fully qualified name by default
- better handling of nested classes via `__qualname__`
- add test
- add type annotations, better doc

before:

`
{'queue': CommandQueue, 'shape': tuple(int,int), 'dtype': dtype[float64], 'strides': tuple(int,int), 'events': list(Event), 'nbytes': int, 'size': int, 'allocator': SVMPool, 'base_data': PooledSVM, 'offset': int, 'context': Context, '_flags': NoneType, 'tags': frozenset(), 'axes': tuple(Axis,Axis)}
`

after:

`
{'queue': pyopencl._cl.CommandQueue, 'shape': tuple(int,int), 'dtype': dtype[float64], 'strides': tuple(int,int), 'events': list(pyopencl._cl.Event), 'nbytes': int, 'size': int, 'allocator': pyopencl._cl.SVMPool, 'base_data': pyopencl._cl.PooledSVM, 'offset': int, 'context': pyopencl._cl.Context, '_flags': NoneType, 'tags': frozenset(), 'axes': tuple(arraycontext.impl.pyopencl.taggable_cl_array.Axis,arraycontext.impl.pyopencl.taggable_cl_array.Axis)}
`

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -34,7 +34,7 @@ import sys
 import logging
 from typing import (
         Any, Callable, Dict, Generic, Hashable, Iterable,
-        List, Optional, Set, Tuple, TypeVar, Union, ClassVar)
+        List, Optional, Set, Tuple, TypeVar, Union, ClassVar, Type, Mapping)
 import builtins
 
 import math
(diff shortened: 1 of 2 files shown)


matthiasdiener:
This is ready for a first look @inducer

pytools/__init__.py
```diff
@@ -2041,12 +2074,12 @@ def typedump(val, max_seq=5, special_handlers=None):
...
+                        objname(val),
                         ",".join(typedump(x, max_seq, special_handlers)
                             for x in val[:max_seq]))
             else:
                 return "{}({})".format(
-                        type(val).__name__,
+                        objname(val),
                         ",".join(typedump(x, max_seq, special_handlers)
                             for x in val))
         except TypeError:
```
matthiasdiener, line 2085: This could probably be changed to `objname(val)` as well, but I couldn't come up with an example that hits this exception here.

inducer:
I think I like these changes. The format strings could probably be f-strings, but otherwise this LGTM.

pytools/__init__.py
matthiasdiener: changed in [7c43184](https://github.com/inducer/pytools/pull/174/commits/7c43184dd3d0d28f6956e94616014c5c405fb00f)

matthiasdiener:
> The format strings could probably be f-strings

I tried to tackle those in [07ba335](https://github.com/inducer/pytools/pull/174/commits/07ba335245702d9bc71a1389d7753b2f367d612c), but I'm not sure they look better than before.

PR #179, opened 2023-06-18
persistent_dict: add support for numpy 1.25

alexfikl:
All the dtypes have moved to `numpy.dtypes` and have names like `Float32DType`. This updates the heuristic to also include this case.

See the release notes https://numpy.org/doc/stable/release/1.25.0-notes.html#numpy-now-exposes-dtype-classes-in-np-dtypes

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -236,12 +236,21 @@ class KeyBuilder:
             tname = tp.__name__
             method = None
             try:
-                method = getattr(self, "update_for_"+tname)
+                method = getattr(self, f"update_for_{tname}")
             except AttributeError:
-                # Handling numpy >= 1.20, for which
-                # type(np.dtype("float32")) -> "dtype[float32]"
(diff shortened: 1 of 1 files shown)


inducer:
Argh. Looks like we unwittingly raced each other on this, see #180. I'll go with mine since it has some extra stuff I had sitting around.

inducer:
Thanks for working on this though, still!

inducer:
Still, it's... possibly reassuring that we both went with approximately the same craptastic approach. :slightly_smiling_face: i

alexfikl:
> Argh. Looks like we unwittingly raced each other on this, see #180. I'll go with mine since it has some extra stuff I had sitting around.

Haha, no worries! That looks good to me (since it's the same fix :grin:).

There's a `pytato` failure because of this too, which I didn't get to
https://github.com/inducer/pytools/actions/runs/5303975533/jobs/9599964177?pr=179

PR #184, opened 2023-09-25
persistent_dict: add complex hashing, numpy scalar hashing

matthiasdiener:
_Please squash_

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -245,10 +245,21 @@ class KeyBuilder:
                         # Handling numpy >= 1.25, for which
                         # type(np.dtype("float32")) -> "Float32DType"
                         or tname.endswith("DType")
+                        or tname in ("int8", "int16", "int32", "int64")
+                        or tname in ("float16", "float32", "float64", "float128")
+                        or tname in ("complex64", "complex128", "complex256")
                         ) and "numpy" in sys.modules:
                     import numpy as np
                     if isinstance(key, np.dtype):
                         method = self.update_for_specific_dtype
+                    elif np.isscalar(key):
+                        # Non-numpy scalars are handled above in the try block.
+                        if np.issubdtype(tp, np.complexfloating):
+                            method = self.update_for_complex
+                        elif np.issubdtype(tp, np.floating):
+                            method = self.update_for_float
+                        elif np.issubdtype(tp, np.integer):
+                            method = self.update_for_int
 
                 elif issubclass(tp, Enum):
                     method = self.update_for_enum
@@ -301,6 +312,10 @@ class KeyBuilder:
                 return
             except OverflowError:
                 sz *= 2
+            except AttributeError:
+                # Numpy scalars don't have to_bytes()
+                key_hash.update(str(key).encode("utf8"))
+                return
 
     @classmethod
     def update_for_enum(cls, key_hash, key):
@@ -312,7 +327,15 @@ class KeyBuilder:
 
     @staticmethod
     def update_for_float(key_hash, key):
-        key_hash.update(key.hex().encode("utf8"))
+        try:
+            key_hash.update(key.hex().encode("utf8"))
+        except AttributeError:
+            # Some numpy float scalars don't have hex()
+            key_hash.update(str(key).encode("utf8"))
+
+    @staticmethod
+    def update_for_complex(key_hash, key):
+        key_hash.update(f"{key.real}.{key.imag}".encode("utf-8"))
 
     @staticmethod
     def update_for_str(key_hash, key):
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -363,9 +363,10 @@ class KeyBuilder:
     def update_for_specific_dtype(key_hash, key):
         key_hash.update(key.str.encode("utf8"))
 
-    def update_for_numpy_scalar(self, key_hash, key):
-        self.rec(key_hash, str(key))
-        self.rec(key_hash, key.dtype)
+    @staticmethod
+    def update_for_numpy_scalar(key_hash, key):
+        import numpy as np
+        key_hash.update(np.array(key).tobytes())
 
     def update_for_dataclass(self, key_hash, key):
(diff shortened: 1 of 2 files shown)


matthiasdiener:
This is ready for review.

pytools/persistent_dict.py
```diff
@@ -301,6 +315,10 @@ def update_for_int(key_hash, key):
                 return
             except OverflowError:
                 sz *= 2
+            except AttributeError:
+                # Numpy scalars don't have to_bytes()
```
inducer, line 319: I think a better approach would be to have all numpy scalars go to the same method, in which you convert to shape-`()` array and then use `to_bytes` on that.

pytools/persistent_dict.py
```diff
@@ -350,6 +363,11 @@ def update_for_dtype(key_hash, key):
     def update_for_specific_dtype(key_hash, key):
         key_hash.update(key.str.encode("utf8"))
 
+    @staticmethod
+    def update_for_numpy_scalar(key_hash, key):
+        import numpy as np
+        key_hash.update(np.atleast_1d(key).tobytes())
```
inducer, line 369: Why go to 1D? Scalars should be 0D.

inducer: `np.array(scalar)` should do the trick.

pytools/persistent_dict.py
matthiasdiener: Should we add similar functionality for ndarrays as well? maybe something like (tobytes, shape, dtype)?

inducer: No, definitely not!
- They're mutable.
- With that implementation, it would violate the fundamental property a hash satisfies, i.e. `a == b` imples `hash(a) == hash(b)`. That's because the array's byte pattern depends on storage order (e.g. F/C ordering), for arrays that Numpy would otherwise consider equal.

matthiasdiener: It seems that `tobytes()` for complex256 (only that dtype) isn't stable:

```python
import numpy as np

for i in range(10):
    k = np.complex256(1.1+2.2j)
    print(k.tobytes())
```

```
$ python t.py
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x00\x00\x00\x00\x00\x00\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@\x00\x00\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x86\xc5\xeel\x16\xfe\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@`w\x00\x00\x00\x00'
b'\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\xff?\x86\xc5\xeel\x16\xfe\x00\xd0\xcc\xcc\xcc\xcc\xcc\x8c\x00@`w\x00\x00\x00\x00'
```

inducer: That makes a little bit of sense to me, as (at least on x86), `float128` (where `complex256` is just two of those) is just the x87 (not a typo) 80-bit FP type. So there are six bytes that aren't accounted for and possibly arbitrary.

Maybe specially handle those? (Cast to `float` (yes, Python `float`, both parts for complex), call `repr` and hash that)?

matthiasdiener: > That makes a little bit of sense to me, as (at least on x86), `float128` (where `complex256` is just two of those) is just the x87 (not a typo) 80-bit FP type. So there are six bytes that aren't accounted for and possibly arbitrary.

Oh 😞  In other words, this might also affect `float128` itself?

> Maybe specially handle those? (Cast to `float` (yes, Python `float`, both parts for complex), call `repr` and hash that)?

If you prefer that idea to https://github.com/inducer/pytools/pull/184/commits/f1b766c791299b08bf4f7090b9b47b4dbfe2678a, I can implement that.

matthiasdiener: I implemented your suggestion in https://github.com/inducer/pytools/pull/184/commits/aeb824ad94a4032d64acfecd3affe7edcb7a88b7 (using Python's `complex()` for converting `np.complex256` because that seemed a bit simper).

inducer:
Thanks!

PR #187, opened 2023-10-31
Add `unique` function.

matthiasdiener:
*Please squash*

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -36,8 +36,8 @@ import sys
 from functools import reduce, wraps
 from sys import intern
 from typing import (
-    Any, Callable, ClassVar, Dict, Generic, Hashable, Iterable, List, Mapping,
-    Optional, Set, Tuple, Type, TypeVar, Union)
+    Any, Callable, ClassVar, Dict, Generic, Hashable, Iterable, Iterator, List,
+    Mapping, Optional, Sequence, Set, Tuple, Type, TypeVar, Union)
 
 
 try:
@@ -189,6 +189,11 @@ String utilities
 ----------------
 
 .. autofunction:: strtobool
+
+Sequence utilities
+------------------
+
+.. autofunction:: unique
 
 Type Variables Used
 -------------------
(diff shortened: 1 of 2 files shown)


inducer:
https://docs.python.org/3/library/itertools.html#itertools.groupby?

matthiasdiener:
> https://docs.python.org/3/library/itertools.html#itertools.groupby?

As far as I can see, `groupby` only removes *consecutive* duplicates, while `unique` removes all duplicates:

```python
>>> from itertools import groupby
>>> [k for k, g in groupby('AAAABBBCCDAABBB')]
['A', 'B', 'C', 'D', 'A', 'B']

>>> from pytools import unique
>>> list(unique('AAAABBBCCDAABBB'))
['A', 'B', 'C', 'D']
```

pytools/__init__.py
```diff
@@ -2994,6 +2999,15 @@ def strtobool(val: Optional[str], default: Optional[bool] = None) -> bool:
 # }}}
 
 
+# {{{ unique
+
+def unique(seq: Sequence[T]) -> Iterator[T]:
+    """Yield unique elements in *seq*. See also :func:`reversed`."""
+    yield from list(dict.fromkeys(seq))
```
inducer:
```suggestion
    return iter(dict.fromkeys())
```

inducer: Saw your simplification: `yield from` is not zero cost, `return` is.

inducer:
> As far as I can see, `groupby` only removes _consecutive_ duplicates, while `unique` removes all duplicates:

That's a fair point. Maybe emphasize that in the docstring?

pytools/__init__.py
matthiasdiener: > Saw your simplification: `yield from` is not zero cost, `return` is.

Thanks, `return iter` is indeed much faster. Done in [d4f9ab3](https://github.com/inducer/pytools/pull/187/commits/d4f9ab3135ddde4ccc1810041b9770865fdb6ade)

matthiasdiener:
> > As far as I can see, `groupby` only removes _consecutive_ duplicates, while `unique` removes all duplicates:
> 
> That's a fair point. Maybe emphasize that in the docstring?

What do you think of [540e565](https://github.com/inducer/pytools/pull/187/commits/540e565fe6e4e11055b529a469b6e102afb10479)?

inducer:
LGTM, thanks! Do you need that released?

PR #191, opened 2023-11-22
remove pytools.log reference from README

matthiasdiener:
(this has been moved to logpyle, as the comment a bit lower indicates)


inducer:
Thx

PR #193, opened 2024-01-31
KeyBuilder: allow hashing of classes that define update_persistent_hash

matthiasdiener:
The three test classes (not objects) are unhashable without this PR.


inducer:
LGTM, thanks!

PR #196, opened 2024-02-19
KeyBuilder: fix hashing of dataclasses/attrs, add test

matthiasdiener:
The test fails without this PR. Followup of #147.

_Please squash_

### DIFF
--- a/pytools/test/test_persistent_dict.py
+++ b/pytools/test/test_persistent_dict.py
@@ -518,6 +518,50 @@ def test_class_hashing():
     assert keyb(TagClass2) != keyb(TagClass2())
 
 
+def test_dataclass_hashing():
+    keyb = KeyBuilder()
+
+    @dataclass
+    class MyDC:
+        name: str
+        value: int
+
+    assert keyb(MyDC("hi", 1)) == \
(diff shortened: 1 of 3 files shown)


pytools/test/test_persistent_dict.py
```diff
@@ -518,6 +518,51 @@ class TagClass2(Tag):
...
+
+    # Class types must be encoded in hash
+    assert keyb(MyDC2("hi", 1)) != keyb(MyDC("hi", 1))
+
+    try:
+        import attrs
+    except ImportError:
+        from warnings import warn
+        warn("attrs not installed, skipping rest of the test")
+        return
```
matthiasdiener, line 545: Not sure what's better here, `warn` or `pytest.importorskip`.

inducer: Break the `attrs` bit out into a separate test and use `importorskip`.

inducer (approved):
LGTM other than the one mini wrinkle.

pytools/test/test_persistent_dict.py
matthiasdiener: Done in 4a1ac64

PR #197, opened 2024-02-19
improve KeyBuilder documentation

matthiasdiener:
Edit: What do you think of renaming the class to `PersistentKeyBuilder` or `PersistentHashBuilder`? We could keep the old name as an alias. This could clarify the goal of the KeyBuilder, as well as its connection to `PersistentDict`.

_Please squash_

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -178,6 +178,23 @@ class KeyBuilder:
     """A (stateless) object that computes hashes of objects fed to it. Subclassing
     this class permits customizing the computation of hash keys.
 
+    Objects of this class are used to calculate hashes of the keys in a
+    :class:`PersistentDict`.
+
+    This class follows the same general rules as Python's built-in hashing:
+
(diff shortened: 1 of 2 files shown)


pytools/persistent_dict.py
```diff
@@ -178,6 +178,23 @@ class KeyBuilder:
...
+    - Objects with the same hash may or may not compare equal.
+
+    In contrast to Python's built-in hashing, using a :class:`KeyBuilder` has
+    several advantages:
+
+    - The hash is persistent across interpreter invocations.
+    - The hash is the same across different Python versions and platforms.
+    - No need to set ``PYTHONHASHSEED`` to get stable hashes for strings/bytes.
+
+
```
matthiasdiener, line 197: Other things we could add here: 
- default type of hash is `str`, not `int`

matthiasdiener: Done in f96ffc7130a847348c4dfbebddd14bcb1098a89c

PR #198, opened 2024-02-20
Tag: remove update_persistent_hash

matthiasdiener:
Needs #196.

Hashing is now handled directly through the dataclass.


inducer:
Thx!

PR #199, opened 2024-02-27
Record: make field ordering deterministic, add test

matthiasdiener:
This is a proof-of-concept and is ready for a first look @inducer. I'm not sure if there is a better way to do this.

_Please squash_

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -473,8 +473,6 @@ class RecordWithoutPickling:
 
 
 class Record(RecordWithoutPickling):
-    __slots__: ClassVar[List[str]] = []
-
     def __getstate__(self):
         return {
                 key: getattr(self, key)
@@ -499,9 +497,6 @@ class Record(RecordWithoutPickling):
             return True
         return (self.__class__ == other.__class__
                 and self.__getstate__() == other.__getstate__())
-
-    def __ne__(self, other):
-        return not self.__eq__(other)
 
 
 class ImmutableRecordWithoutPickling(RecordWithoutPickling):
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -465,7 +465,7 @@ class RecordWithoutPickling:
         except AttributeError:
             self.__class__.fields = fields = {}
 
-        fields.update(dict.fromkeys(new_fields))
+        fields.update(dict.fromkeys(sorted(new_fields)))
 
     def __getattr__(self, name):
         # This method is implemented to avoid pylint 'no-member' errors for
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -417,7 +417,8 @@ class RecordWithoutPickling:
 
     def __init__(self, valuedict=None, exclude=None, **kwargs):
         from warnings import warn
-        warn("pytools.Record is deprecated. Use dataclasses instead.")
+        warn(f"{self.__class__.__bases__[0]} is deprecated and will be "
+             "removed in 2025. Use dataclasses instead.")
 
         assert self.__class__ is not Record
 
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -411,7 +411,9 @@ class RecordWithoutPickling:
     """
 
     __slots__: ClassVar[List[str]] = []
-    fields: ClassVar[Set[str]]
+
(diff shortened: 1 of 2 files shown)


inducer:
- This seems OK, although it is a compatibility break, which I'm not loving.
- We should just deprecate the thing for removal in 2025. I.e. have it spew warnings, and remove it from the docs. (in favor of dataclasses, which do the same thing, but more competently)

matthiasdiener:
> * This seems OK, although it is a compatibility break, which I'm not loving.

Is there a use case that breaks with this change? My goal was to remain compatible with the previous set-based implementation (which is the reason for the conversion from set to dict in `__init__`).

> * We should just deprecate the thing for removal in 2025. I.e. have it spew warnings, and remove it from the docs. (in favor of dataclasses, which do the same thing, but more competently)

What do you think of [c32e6d0](https://github.com/inducer/pytools/pull/199/commits/c32e6d01def6383a126059a9932f0c320783c7a1)?

matthiasdiener:
Setting this as ready for review. mirgecom seems to run fine with it.

pytools/__init__.py
```diff
@@ -468,8 +476,6 @@ def __getattr__(self, name):
 
 
 class Record(RecordWithoutPickling):
-    __slots__: ClassVar[List[str]] = []
```
inducer, line 471: Is there a specific reason this was dropped?

pytools/__init__.py
```diff
@@ -411,9 +411,14 @@ class RecordWithoutPickling:
...
 
     __slots__: ClassVar[List[str]] = []
-    fields: ClassVar[Set[str]]
+
+    # A dict, not a set, to maintain a deterministic iteration order
+    fields: ClassVar[Dict[str, None]]
 
     def __init__(self, valuedict=None, exclude=None, **kwargs):
+        from warnings import warn
+        warn("pytools.Record is deprecated. Use dataclasses instead.")
```
inducer, line 420: Also remove from docs.

inducer, line 420: Note that this is `RecordWithoutPickling`.

matthiasdiener: Is it documented somewhere? I could not find it.

inducer: Looks like we already got rid of it. Good! Nevermind then. :slightly_smiling_face:

pytools/__init__.py
matthiasdiener: I had overlooked that this must be defined in every subclass to be effective 🤦 . Restored in [0a0fcef](https://github.com/inducer/pytools/pull/199/commits/0a0fcef799d5fc19e1e27541b298d524943197a1)

pytools/__init__.py
matthiasdiener: Right, would you prefer something like `pytools.{self.__class__}` or `{pytools.{self.__class__.__bases__[0]}}`  ?

pytools/__init__.py
```diff
@@ -411,9 +411,16 @@ class RecordWithoutPickling:
     """
 
     __slots__: ClassVar[List[str]] = []
-    fields: ClassVar[Set[str]]
 
-    def __init__(self, valuedict=None, exclude=None, **kwargs):
+    # A dict, not a set, to maintain a deterministic iteration order
+    fields: ClassVar[Dict[str, None]]
```
matthiasdiener, line 416: Side note: this change will require mypy-related changes in loopy

pytools/__init__.py
inducer: Yes, probably. But :shrug: not urgent, we're getting rid of this anyway.

inducer:
I'll be honest, I underestimated the compatibility impact of this. I've just pushed a revert.

inducer:
https://github.com/inducer/loopy/pull/785 gets rid of a good bunch more `Record`s in Loopy, which may make this more plausible.

matthiasdiener:
Sure, no problem! Sorry for the confusion. Is there a particular error that came up with this?

inducer:
It was code like this that made me substantially uneasy:

https://github.com/inducer/loopy/blob/d5512d5c6aec12f16a60aa5da31701c23d006bbd/loopy/kernel/instruction.py#L206-L211

PR #200, opened 2024-03-12
Guard deprecated usage of longfloat and longcomplex

alexfikl:
Found using `ruff check --select NPY201 .` as suggested in the numpy 2.0 migration guide.


inducer:
Might as well remove these outright.

alexfikl:
> Might as well remove these outright.

@inducer Sure thing! They're removed now.

PR #201, opened 2024-03-14
PersistentDict: compress pickled data

matthiasdiener:


pytools/persistent_dict.py
inducer, line 1: If we're unconditionally compressing, then we should bump the on-disk format version.

matthiasdiener: ie, it should become `pdict-v5-...`?

inducer: Yes.

matthiasdiener: Done in https://github.com/inducer/pytools/pull/201/commits/5308f332b49c6c2aa27641f1e30c365f8c6a6458

PR #206, opened 2024-03-20
Bump version to 2024.1

matthiasdiener:
(perhaps it should be 2024.2 now since there have been new commits to `main` since the [pypi release](https://pypi.org/project/pytools/2024.1/#files)?)


inducer:
I got halfway there yesterday, releasing 2024.1 to pypi before I got interrupted. There's no tag yet. I'll clean up the mess now.

inducer:
2024.1.1 is out. It should address both aspects here.

PR #208, opened 2024-03-21
[proof-of-concept] PersistentDict: Use sqlite as backend storage

matthiasdiener:
Pros
- Should be able to handle concurrent access natively (https://www.sqlite.org/faq.html#q5)
- Much simpler code

Cons:
- No way to check for hash collisions
- Could probably benefit from using `sqlite3` directly instead of going through `sqlitedict`
  - `sqlitedict` has multithreading support that we don't need
  - could optimize queries, e.g. in the `WriteOncePersistentDict.store` implementation)

TODOs:
- [x] close DB

References:
- https://github.com/litements/litedict
- https://github.com/piskvorky/sqlitedict


inducer:
Superseded by #209, closing.

PR #209, opened 2024-03-27
PersistentDict: Use sqlite as backend storage v2

matthiasdiener:
Based on https://github.com/matthiasdiener/skvlite (which itself is based on https://github.com/litements/litedict).

Alternative to #208.

TODOs:
- [x] type annotations
- [x] better doc
- [x] derive from `Mapping` (851905a)
- [x] performance tests
- [x] storage test
- ~~compression (maybe as part of a different PR)~~

----

**Performance**

`time -p pytest -sk test_speed`, Mac M1

| Config                                         | write (s) | read (s) | time -p (s) |
|------------------------------------------------|-----------|----------|-------------|
| Baseline (`main`)                              | 3.35      | 2.71     | 9.92        |
| WAL=OFF, SYNC=NORMAL (this PR's default config)| 3.70      | 0.42     | 4.67        |
| WAL=OFF, SYNC=OFF                              | 2.73      | 0.42     | 3.79        |
| WAL=ON, SYNC=NORMAL                            | 0.22      | 0.23     | 0.90        |

At least with the current implementation, WAL=ON isn't safe when multiple processes access the database, as indicated by the loopy CI failure: https://github.com/inducer/pytools/actions/runs/8469347803/job/23204482712. It works only on non-networked filesystems. See also https://www.sqlite.org/wal.html.

---

**Storage size**

`pytest -k test_size`, Mac M1

| Config | Method | Size |
|-------------|--------|------|
| Baseline (`main`) | `du -sh` | 79 MByte |
| With this PR (saving `keyhash, (key, value)`) | `.nbytes()` | 1.88 MByte |
| With earlier version of this PR (saving `keyhash, value`) | `.nbytes()` | 1.64 MByte |

---

_Please squash_

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -343,9 +343,20 @@ class NoSuchEntryError(KeyError):
     pass
 
 
+class NoSuchEntryCollisionError(NoSuchEntryError):
+    """Raised when an entry is not found in a :class:`PersistentDict`, but it
+    contains an entry with the same hash key (hash collision)."""
+    pass
+
+
 class ReadOnlyEntryError(KeyError):
     """Raised when an attempt is made to overwrite an entry in a
     :class:`WriteOncePersistentDict`."""
+    pass
+
+
+class CollisionWarning(UserWarning):
+    """Warning raised when a collision is detected in a :class:`PersistentDict`."""
     pass
 
 
@@ -408,6 +419,24 @@ class _PersistentDictBase:
         if self.conn:
             self.conn.close()
 
+    def _collision_check(self, key, stored_key):
+        if stored_key != key:
+            print(stored_key, key)
+            # Key collision, oh well.
+            from warnings import warn
+            warn(f"{self.identifier}: key collision in cache at "
+                    f"'{self.container_dir}' -- these are sufficiently unlikely "
+                    "that they're often indicative of a broken hash key "
+                    "implementation (that is not considering some elements "
+                    "relevant for equality comparison)",
+                    CollisionWarning
+                 )
+
+            # This is here so we can step through equality comparison to
+            # see what is actually non-equal.
+            stored_key == key  # pylint:disable=pointless-statement  # noqa: B015
+            raise NoSuchEntryCollisionError(key)
+
     def store_if_not_present(self, key: Any, value: Any) -> None:
         """Store (*key*, *value*) if *key* is not already present."""
         self.store(key, value, _skip_if_present=True)
@@ -448,12 +477,12 @@ class _PersistentDictBase:
     def values(self) -> Generator[Any, None, None]:
         """Return an iterator over the values in the dictionary."""
         for row in self.conn.execute("SELECT value FROM dict ORDER BY rowid"):
-            yield pickle.loads(row[0])
+            yield pickle.loads(row[0])[1]
 
     def items(self) -> Generator[tuple[str, Any], None, None]:
         """Return an iterator over the items in the dictionary."""
         for row in self.conn.execute("SELECT key, value FROM dict ORDER BY rowid"):
-            yield (row[0], pickle.loads(row[1]))
+            yield (row[0], pickle.loads(row[1])[1])
 
     def size(self) -> int:
         """Return the size of the dictionary in bytes."""
@@ -511,18 +540,18 @@ class WriteOncePersistentDict(_PersistentDictBase):
         self._fetch.cache_clear()
 
     def store(self, key: Any, value: Any, _skip_if_present: bool = False) -> None:
-        k = self.key_builder(key)
-        v = pickle.dumps(value)
+        keyhash = self.key_builder(key)
+        v = pickle.dumps((key, value))
 
         try:
-            self.conn.execute("INSERT INTO dict VALUES (?, ?)", (k, v))
+            self.conn.execute("INSERT INTO dict VALUES (?, ?)", (keyhash, v))
         except sqlite3.IntegrityError:
             if not _skip_if_present:
                 raise ReadOnlyEntryError("WriteOncePersistentDict, "
                                          "tried overwriting key")
 
     def _fetch(self, keyhash: str) -> Any:  # pylint:disable=method-hidden
-        # This is separate from fetch() to allow for LRU caching
+        # This method is separate from fetch() to allow for LRU caching
         c = self.conn.execute("SELECT value FROM dict WHERE key=?", (keyhash,))
         row = c.fetchone()
         if row is None:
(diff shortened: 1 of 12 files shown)


matthiasdiener:
This is ready for a first look @inducer.

pytools/persistent_dict.py
```diff
@@ -449,38 +343,18 @@ class NoSuchEntryError(KeyError):
...
-    """Raised when an entry is not found in a :class:`PersistentDict` due to an
-    invalid contents file."""
-    pass
-
-
-class NoSuchEntryCollisionError(NoSuchEntryError):
-    """Raised when an entry is not found in a :class:`PersistentDict`, but it
-    contains an entry with the same hash key (hash collision)."""
-    pass
-
```
inducer, line 468: Existing code may import these, so we're not really at liberty to remove them. Why are you proposing to remove them?

pytools/test/test_persistent_dict.py
```diff
@@ -202,37 +202,6 @@ def test_persistent_dict_synchronization():
         shutil.rmtree(tmpdir)
 
 
-def test_persistent_dict_cache_collisions():
```
inducer, line 205: For each deleted test, could you explain why they're no longer relevant?

pytools/persistent_dict.py
matthiasdiener: `NoSuchEntryInvalidKeyError` and `NoSuchEntryInvalidContentsError` can't occur anymore, because we don't have separate key and content files with this PR. We could provide a custom exception for something like 'invalid sqlite file', but I'm not sure that would be more helpful for the application.

Regarding `NoSuchEntryCollisionError`, we currently have no way of checking for collisions, since the key itself is not stored in the sqlite file, just its hash.

inducer:
As an overall comment: Without cache collision checking, this is an immediate nonstarter to me, because those are very real bugs that occur from time to time that are an immense pain to debug unless the infrastructure catches them before they become bugs.

That said, I agree that separately storing the key is often duplicative and expensive. A simple way to help mitigate that expense would be to store a `(key, value)` tuple in a single tuple. Then any repeated objects between the two (which I expect will be many) are stored essentially for free.

pytools/test/test_persistent_dict.py
matthiasdiener: 367683c & 7028e6b restore the collision tests.

pytools/persistent_dict.py
matthiasdiener: https://github.com/inducer/pytools/commit/367683c6eaac0b0cc2a6048d3de5b30d45a0605a & https://github.com/inducer/pytools/commit/7028e6bbd63ef734e5d23cb8b144f06dcb2153c2 restore the collision tests.

matthiasdiener:
> That said, I agree that separately storing the key is often duplicative and expensive. A simple way to help mitigate that expense would be to store a `(key, value)` tuple in a single tuple. Then any repeated objects between the two (which I expect will be many) are stored essentially for free.

Thanks, I implemented your suggestion in https://github.com/inducer/pytools/commit/367683c6eaac0b0cc2a6048d3de5b30d45a0605a.

matthiasdiener:
This is ready for another review @inducer.

matthiasdiener:
@inducer , what do you think of potentially splitting this from pytools and create a new package (similar to logpyle)? I have not been able to find any reasonable other implementation of a key-value store in Python (due to, in part, the extensive hashing functionality here).

pytools/persistent_dict.py
```diff
@@ -549,112 +447,122 @@ def __init__(self, identifier: str,
...
-        """Create the container directory to store the dictionary."""
-        os.makedirs(self.container_dir, exist_ok=True)
+    def __del__(self) -> None:
+        if self.conn:
+            self.conn.close()
 
-    def _collision_check(self, key: K, stored_key: K, _stacklevel: int) -> None:
+    def _collision_check(self, key: K, stored_key: K) -> None:
         if stored_key != key:
+            print(stored_key, key)
```
inducer, line 491:
```suggestion
```

pytools/persistent_dict.py
```diff
@@ -549,112 +447,122 @@ def __init__(self, identifier: str,
...
+        """Return an iterator over the values in the dictionary."""
+        for row in self.conn.execute("SELECT key_value FROM dict ORDER BY rowid"):
+            yield pickle.loads(row[0])[1]
+
+    def items(self) -> Generator[tuple[K, V], None, None]:
+        """Return an iterator over the items in the dictionary."""
+        for row in self.conn.execute("SELECT key_value FROM dict ORDER BY rowid"):
+            yield pickle.loads(row[0])
+
+    def size(self) -> int:
```
inducer, line 554:
```suggestion
    def nbytes(self) -> int:
```
?

pytools/persistent_dict.py
```diff
@@ -698,129 +610,38 @@ def clear_in_mem_cache(self) -> None:
...
-
-        # {{{ load contents
+            self.conn.execute("INSERT INTO dict VALUES (?, ?)", (keyhash, v))
+        except sqlite3.IntegrityError:
+            if not _skip_if_present:
+                raise ReadOnlyEntryError("WriteOncePersistentDict, "
+                                         "tried overwriting key")
+
+    def _fetch(self, keyhash: str) -> Tuple[K, V]:  # pylint:disable=method-hidden
+        # This method is separate from fetch() to allow for LRU caching
```
inducer, line 627: Does the separation still make sense? There's no more LRU AFAICT. At least this comment can probably go.

pytools/persistent_dict.py
```diff
@@ -549,112 +447,122 @@ def __init__(self, identifier: str,
...
+        # https://www.sqlite.org/wal.html
+        if enable_wal:
+            self.conn.execute("PRAGMA journal_mode = 'WAL'")
 
-    def _key_file(self, hexdigest_key: str) -> str:
-        from os.path import join
-        return join(self._item_dir(hexdigest_key), "key")
+        # temp_store=2: use in-memory temp store
+        # https://www.sqlite.org/pragma.html#pragma_temp_store
+        self.conn.execute("PRAGMA temp_store = 2")
```
inducer, line 476: Could you roughly describe the findings that led to each of these?

pytools/persistent_dict.py
```diff
@@ -549,112 +447,122 @@ def __init__(self, identifier: str,
...
+        # https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions
+        self.conn = sqlite3.connect(self.filename, isolation_level=None)
 
-        # Some file systems limit the number of directories in a directory.
-        # For ext4, that limit appears to be 64K for example.
-        # This doesn't solve that problem, but it makes it much less likely
+        self.conn.execute(
+            "CREATE TABLE IF NOT EXISTS dict "
+            "(keyhash TEXT NOT NULL PRIMARY KEY, key_value TEXT NOT NULL)"
+            )
```
inducer, line 468: Explain somewhere that it's a deliberate choice to not store all values with a certain key, based on the assumption that key conflicts are supremely unlikely, and if they occur, almost always due to a bug.

inducer (commented):
This looks good! Just a few minor wrinkles.

pytools/persistent_dict.py
matthiasdiener: LRU should still be there, see line 605 in this file.

pytools/persistent_dict.py
matthiasdiener: I added a comment regarding this in 51666d0.

pytools/persistent_dict.py
matthiasdiener: What do you think of b4b623a?

inducer:
Thx!

PR #212, opened 2024-03-28
PersistentDict: better docs, type annotations

matthiasdiener:
(extracted from #209)

_Please squash_

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -75,6 +75,18 @@ This module also provides a disk-backed dictionary that uses persistent hashing.
 .. autoclass:: KeyBuilder
 .. autoclass:: PersistentDict
 .. autoclass:: WriteOncePersistentDict
+
+
+Internal stuff that is only here because the documentation tool wants it
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. class:: K
+
+    A type variable for the key type of a :class:`PersistentDict`.
+
+.. class:: V
+
+    A type variable for the value type of a :class:`PersistentDict`.
 """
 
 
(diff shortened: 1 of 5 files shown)


pytools/persistent_dict.py
```diff
@@ -479,7 +479,9 @@ class CollisionWarning(UserWarning):
 
 
 class _PersistentDictBase:
```
inducer, line 481: Conceivably, we should make this a generic in terms of its key and value types.

matthiasdiener: Should this derive from `MutableMapping`?

inducer: Doesn't mypy figure out structurally whether something is a mapping?

matthiasdiener: I'm not sure. I just realized that `Mapping` might not make too much sense here, since it the classes don't have `keys()` etc. It might make more sense in the context of #209.

matthiasdiener: Is [f95b64c](https://github.com/inducer/pytools/pull/212/commits/f95b64c5a99f6fb9811cd38b30ad9016bc6426d2) what you had in mind? I couldn't find a way to make `WriteOnce...` generic as well, because then all consumers also would need to change their type annotations.

inducer:
Could you look at the docs CI failures?

matthiasdiener:
> Could you look at the docs CI failures?

Fixed in [ca19deb](https://github.com/inducer/pytools/pull/212/commits/ca19debd532b49ee86fd8de0364cbb288d2db0c9)

pytools/persistent_dict.py
inducer: Please just go ahead. I'm not worried about breaking downstream type annotations. They can always be fixed, and they don't prevent anyone from doing their work. At least mostly.

matthiasdiener: What do you think of 2a8e506?

inducer:
LGTM, thanks!

PR #214, opened 2024-04-01
type-annotate Record

matthiasdiener:


matthiasdiener:
> Don't know. We're trying to get rid of it. What's the benefit of annotating it?

Not much, but is it used in `DataTable` (and prevents its full annotation).

inducer:
Eh, OK. Doesn't hurt. Could you look into the mypy failure?

PR #216, opened 2024-04-12
add deprecation_warning_on_import

matthiasdiener:
(could be useful in the context of #209 when somebody imports `NoSuchEntryInvalidKeyError` or `NoSuchEntryInvalidContentError`)

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -282,6 +282,38 @@ def deprecate_keyword(oldkey: str,
 
     return wrapper
 
+
+from collections import defaultdict
+from warnings import warn
+
+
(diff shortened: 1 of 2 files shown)


matthiasdiener:
this is a proof-of-concept, but ready for a first look @inducer

pytools/__init__.py
inducer, line 1: I think this is way too costly. Imagine this being used in 50 modules. Suddenly, every import has to go through 50 recursive levels. A better way is to use module-level `__getattr__`, as is already being done in a few places.

matthiasdiener: TIL, thanks (fbb985c).

PR #217, opened 2024-04-12
CI: test more Python versions

matthiasdiener:


inducer:
I'm not sure I'm too enthusiastic about updating these once a year. Oldest supported and current seem good enough to me.

matthiasdiener:
Ok, no problem!

PR #218, opened 2024-04-14
clean up some version_info checks

matthiasdiener:
(minimum required version is 3.8 currently)

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -382,9 +382,9 @@ def __getattr__(name):
     if replacement_and_obj is not None:
         replacement, obj, year = replacement_and_obj
         from warnings import warn
-        warn(f"'arraycontext.{name}' is deprecated. "
+        warn(f"'pytools.tag.{name}' is deprecated. "
                 f"Use '{replacement}' instead. "
-                f"'arraycontext.{name}' will continue to work until {year}.",
+                f"'pytools.tag.{name}' will continue to work until {year}.",
                 DeprecationWarning, stacklevel=2)
         return obj
     else:
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -28,7 +28,6 @@ Internal stuff that is only here because the documentation tool wants it
     A type variable with lower bound :class:`Taggable`.
 """
 
-import sys
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


pytools/tag.py
```diff
@@ -378,23 +377,18 @@ class IgnoredForEqualityTag(Tag):
...
-else:
-    TagsType = FrozenSet[Tag]
-    TagOrIterableType = ToTagSetConvertible
-    T_co = TypeVar("TaggableT", bound="Taggable")
+def __getattr__(name):
+    replacement_and_obj = _depr_name_to_replacement_and_obj.get(name, None)
+    if replacement_and_obj is not None:
+        replacement, obj, year = replacement_and_obj
+        from warnings import warn
+        warn(f"'arraycontext.{name}' is deprecated. "
```
matthiasdiener, line 385: should this be

```suggestion
        warn(f"'pytools.tag.{name}' is deprecated. "
```

? (same below)

pytools/test/test_pytools.py
```diff
@@ -31,7 +31,6 @@
 from typing import FrozenSet
 
 
-@pytest.mark.skipif("sys.version_info < (2, 5)")
```
alexfikl, line 34: Did this even work? I thought the first argument had to be a bool.

pytools/tag.py
alexfikl: Yeah, definitely! This was just copy pasted from a similar `arraycontext` change from what I recall.

pytools/test/test_pytools.py
matthiasdiener: I *think* it worked, see e.g. here: https://docs.pytest.org/en/7.1.x/historical-notes.html#string-conditions

pytools/tag.py
matthiasdiener: Changed in [6dd3883](https://github.com/inducer/pytools/pull/218/commits/6dd3883b1edff586e3f8b2bbd7ed06091695d987)

pytools/test/test_pytools.py
alexfikl: Oooh! Neat!

inducer:
Thx!

PR #219, opened 2024-04-16
KeyBuilder: add datetime hashing

matthiasdiener:


pytools/persistent_dict.py
```diff
@@ -449,6 +449,9 @@ def update_for_frozendict(self, key_hash: Hash, key: Mapping) -> None:
     update_for_PMap = update_for_frozendict  # noqa: N815
     update_for_Map = update_for_frozendict  # noqa: N815
 
+    def update_for_datetime(self, key_hash: Hash, key: Any) -> None:
+        self.rec(key_hash, key.isoformat(timespec="microseconds"))
```
inducer, line 453: I don't think this is correct. Two `datetime`s from different timezones can compare equal. Look at what `datetime.__eq__` does? Maybe use `timestamp`? (And please add a test for the timezone thing.)

inducer:
Could you look at the failures here? (Or move this to draft status if it's not ready?)

matthiasdiener:
Thanks @inducer , I think this is ready for another look

inducer:
Thx!

PR #220, opened 2024-04-22
Tag propagation graph generation

a-alveyblanc:
- Moves tag propagation graph generation logic from [pytato](https://github.com/inducer/pytato/pull/494) to pytools

### DIFF
--- a/pytools/graph.py
+++ b/pytools/graph.py
@@ -83,6 +83,7 @@ from typing import (
     Mapping,
     MutableSet,
     Optional,
+    Protocol,
     Set,
     Tuple,
     TypeVar,
@@ -267,8 +268,13 @@ class CycleError(Exception):
         self.node = node
 
 
+class _SupportsLT(Protocol):
+    def __lt__(self, other: object) -> bool:
+        ...
+
+
 @dataclass(frozen=True)
-class HeapEntry(Generic[NodeT]):
+class _HeapEntry(Generic[NodeT]):
     """
     Helper class to compare associated keys while comparing the elements in
     heap operations.
@@ -277,9 +283,9 @@ class HeapEntry(Generic[NodeT]):
     <https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Lib/heapq.py#L135-L138>.
     """
     node: NodeT
-    key: Any
+    key: _SupportsLT
 
-    def __lt__(self, other: HeapEntry) -> bool:
+    def __lt__(self, other: _HeapEntry[NodeT]) -> bool:
         return self.key < other.key
 
 
@@ -325,7 +331,7 @@ def compute_topological_order(graph: GraphT[NodeT],
     # heap: list of instances of HeapEntry(n) where 'n' is a node in
     # 'graph' with no predecessor. Nodes with no predecessors are the
     # schedulable candidates.
-    heap = [HeapEntry(n, keyfunc(n))
+    heap = [_HeapEntry(n, keyfunc(n))
             for n, num_preds in nodes_to_num_predecessors.items()
             if num_preds == 0]
     heapify(heap)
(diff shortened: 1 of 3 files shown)


pytools/graph.py
```diff
@@ -520,5 +521,89 @@ def dfs(node: NodeT) -> None:
 
     return visited == graph.keys()
 
+# }}}
+
+
+# {{{ tag propagation graph processing
+
+def get_propagation_graph_from_constraints(
```
inducer, line 529: Add to docs, explain context a bit, describe the semantics.

pytools/tag.py
```diff
@@ -362,6 +362,18 @@ class IgnoredForEqualityTag(Tag):
 # }}}
 
 
+# {{{ IgnoredForPropagationTag
+
+class IgnoredForPropagationTag(UniqueTag):
+    """
+    Used to signify that a Taggable should *not* have any tags propagated along
```
inducer, line 369: Add some context (in terms of referencing the functions added in this PR) on what "propagation" is.

pytools/tag.py
```diff
@@ -362,6 +362,18 @@ class IgnoredForEqualityTag(Tag):
 # }}}
 
 
+# {{{ IgnoredForPropagationTag
+
+class IgnoredForPropagationTag(UniqueTag):
```
inducer, line 367: We should think about the semantics of this a bit more carefully, I think.

image

There are potentially two "propagation edges" (at least in the pytato context) that could be affected by this, see picture. Which of them is intended to be affected?

We could still represent both types of edges at this level of abstraction, by talking about the "constraint LHS" and "constraint RHS".

pytools/graph.py
```diff
@@ -520,5 +521,89 @@ def dfs(node: NodeT) -> None:
...
+        constraints: List[Tuple[str, str]],
```
inducer, line 530:
```suggestion
        constraints: Sequence[Tuple[str, str]],
```
? Or even just `Iterable`?

pytools/graph.py
```diff
@@ -520,5 +521,89 @@ def dfs(node: NodeT) -> None:
...
+        metadata=None
```
inducer, line 531: Add type for `metadata`.

pytools/graph.py
```diff
@@ -520,5 +521,89 @@ def dfs(node: NodeT) -> None:
...
+                    new_constraints.remove(constraint)
+
+        constraints = new_constraints.copy()
+
+    # }}}
+
+    # {{{ generate propagation graph using constraints
+
+    for lhs, rhs in constraints:
+        assert lhs != rhs
```
inducer, line 575: This is input checking, should not use `assert`.

pytools/graph.py
inducer, line 1: Make sure to preserve copyright. Make sure to add yourself.

a-alveyblanc:
@inducer This is ready for another look when you get the chance

pytools/graph.py
```diff
@@ -47,6 +48,7 @@
 .. autofunction:: as_graphviz_dot
 .. autofunction:: validate_graph
 .. autofunction:: is_connected
+.. autofunction:: get_propagation_graph_from_constraints
```
inducer, line 51: This is still missing `get_reachable_nodes` as well as a (probably longer) narrative on what these functions are even *for*.

pytools/graph.py
```diff
@@ -520,5 +523,61 @@ def dfs(node: NodeT) -> None:
...
     return visited == graph.keys()
 
+# }}}
+
+
+# {{{ tag propagation graph processing
+
+def get_propagation_graph_from_constraints(
+        constraints: Iterable[Tuple[str, str]],
+    ) -> Mapping[str, FrozenSet[str]]:
```
inducer, line 533: Is `str` really essential here? Could this be `GraphNodeT`, so that this could return `GraphT`?

pytools/graph.py
```diff
@@ -76,6 +78,7 @@
 
 
 NodeT = TypeVar("NodeT", bound=Hashable)
+GraphNodeT = TypeVar("GraphNodeT")
```
inducer, line 81: Why is this separate from `NodeT`?

inducer:
Could you look into the linter failures?

pytools/graph.py
```diff
@@ -542,5 +546,71 @@ def dfs(node: NodeT) -> None:
 
     return visited == graph.keys()
 
+# }}}
+
+
+# {{{ tag propagation graph processing
+
+def get_propagation_graph_from_constraints(
```
inducer, line 554:
```suggestion
def undirected_graph_from_edges(
```

inducer:
Thx!

PR #224, opened 2024-05-21
KeyBuilder: support function hashing

matthiasdiener:


pytools/test/test_persistent_dict.py
```diff
@@ -719,6 +719,44 @@ def test_xdg_cache_home() -> None:
...
+    keyb = KeyBuilder()
+
+    assert keyb(global_fun) == keyb(global_fun) == \
+        "51b5980dd3a8aa13f6e83869e4a04c22973d7aaf96cb22899abdfdc55e15c9b2"
+    assert keyb(global_fun) != keyb(global_fun2)
+
+    def local_fun():
+        pass
+
+    def local_fun2():
```
inducer, line 740: I wonder whether we should offer to hash local functions, given that they can be closures. For example:

```py
def get_fun(x):
    def add_x(y):
        return x + y
    return add_x
```

Would `gen_fun(5)` hash equal to `get_fun(7)`? Should it (probably not)?

matthiasdiener: Yeah, they do hash equal currently (but the functions themselves do not compare equal).

matthiasdiener: 6c8a26b improves this a bit, but there is still a small inconsistency (see 6aef85e).

inducer: That's OK IMO. Hashing equal for not-equal objects is OK.

PR #225, opened 2024-05-21
add to_identifier

matthiasdiener:


alexfikl:
I was just reading through some PR emails and was wondering: Is this supposed to ensure that the result can be used as an identifier, e.g. variable name? If yes, I'm not sure this is enough because it can still return keywords and will fail.
```
from keyword import iskeyword
if iskeyword(s):
	s = f"_{s}"
```

matthiasdiener:
> I was just reading through some PR emails and was wondering: Is this supposed to ensure that the result can be used as an identifier, e.g. variable name? If yes, I'm not sure this is enough because it can still return keywords and will fail. Adding something like this could fix that issue (?)
> 
> ```
> from keyword import iskeyword
> if iskeyword(s):
> 	s = f"_{s}"
> ```

My intention for `to_identifier` was that the string that is returned passes the `isidentifier()` test. We could add a new function that also checks for keywords, like you suggested.

alexfikl:
> My intention for `to_identifier` was that the string that is returned passes the `isidentifier()` test. We could add a new function that also checks for keywords, like you suggested.

If there's no need, then no point in making life difficult.. I wanted to mention it, just in case :grin:

Issue #227, opened 2024-05-31
Issues with sqlite `PersistentDict`


inducer:
Now that #209 is in and released, at least two bugs have turned up in it:

- CI fails on Gitlab, apparently after being killed: https://gitlab.tiker.net/inducer/pytools/-/jobs/700185
- (Apparent) concurrent access to a cache in PyOpenCL's CI on Gitlab results in
  `sqlite3.OperationalError: database is locked`: https://gitlab.tiker.net/inducer/pyopencl/-/jobs/700209.

I will go ahead and yank 2024.1.4 because of these issues.

cc @matthiasdiener

matthiasdiener:
> * (Apparent) concurrent access to a cache in PyOpenCL's CI on Gitlab results in
>   `sqlite3.OperationalError: database is locked`: https://gitlab.tiker.net/inducer/pytools/-/jobs/700185.

Do you mean https://gitlab.tiker.net/inducer/pyopencl/-/jobs/700220 ?

inducer:
I meant https://gitlab.tiker.net/inducer/pyopencl/-/jobs/700209, sorry. Edited above.

inducer:
(But they look similar.)

matthiasdiener:
There is something .... strange... going on with the beers (tested on `stout` and `porter`):

On Ubuntu CI (e.g., https://github.com/inducer/pytools/actions/runs/9322161221/job/25662628945):
```
============================= slowest 10 durations =============================
11.70s call     pytools/test/test_persistent_dict.py::test_speed
[...]
```

On porter/stout, that same test takes 800s.

inducer:
26.6s on lager (ext4, RAID0 via Dell controller). Quite a while (hasn't finished yet) on my laptop (ext4, Samsung SSD).

matthiasdiener:
On my Mac M1

> * 26.6s on lager (ext4, RAID0 via Dell controller).
> * Six minutes (356s) on my laptop (ext4, Samsung NVME SSD). `libsqlite3-0 3.45.3-1`
> * 32.4s on the Raspberry Pi 5 that's hosting https://gitlab.tiker.net/ (ext4, WD? NVME SSD) `libsqlite3-0 3.45.3-1`
> 
> Mount options are (mostly?) identical across all three. (`rw,relatime`). Some have `errors=remount-ro`, but I doubt that makes a difference.

Are the `/tmp` filesystems all using ext4?

inducer:
```
[...]

PR #228, opened 2024-06-01
PersistentDict: fixes/workarounds for #227

matthiasdiener:
Should address issues in #227.

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -460,7 +460,8 @@ class _PersistentDictBase(Mapping[K, V]):
 
         # isolation_level=None: enable autocommit mode
         # https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions
-        self.conn = sqlite3.connect(self.filename, isolation_level=None)
+        self.conn = sqlite3.connect(self.filename, isolation_level=None,
+                                    timeout=60)
 
         self.conn.execute(
             "CREATE TABLE IF NOT EXISTS dict "
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -628,12 +628,24 @@ class WriteOncePersistentDict(_PersistentDictBase[K, V]):
         keyhash = self.key_builder(key)
         v = pickle.dumps((key, value))
 
-        try:
-            self.conn.execute("INSERT INTO dict VALUES (?, ?)", (keyhash, v))
-        except sqlite3.IntegrityError:
-            if not _skip_if_present:
-                raise ReadOnlyEntryError("WriteOncePersistentDict, "
-                                         "tried overwriting key")
+        if _skip_if_present:
+            self.conn.execute("INSERT OR IGNORE INTO dict VALUES (?, ?)",
+                              (keyhash, v))
+        else:
+            try:
+                self.conn.execute("INSERT INTO dict VALUES (?, ?)", (keyhash, v))
+            except sqlite3.IntegrityError as e:
+                if hasattr(e, "sqlite_errorcode"):
+                    if e.sqlite_errorcode == 1555:
+                        # 1555 == SQLITE_CONSTRAINT_PRIMARYKEY
+                        # https://sqlite.org/rescode.html#constraint_primarykey
+                        raise ReadOnlyEntryError("WriteOncePersistentDict, "
+                                                 "tried overwriting key")
+                    else:
+                        raise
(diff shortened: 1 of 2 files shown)


matthiasdiener:
Needs more testing on the beers, but otherwise ready for a first look @inducer

pytools/persistent_dict.py
```diff
@@ -460,7 +460,8 @@ def __init__(self, identifier: str,
 
         # isolation_level=None: enable autocommit mode
         # https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions
-        self.conn = sqlite3.connect(self.filename, isolation_level=None)
+        self.conn = sqlite3.connect(self.filename, isolation_level=None,
+                                    timeout=60)
```
alexfikl, line 464: If I understand correctly, the default timeout is 5s and this increases it to 60s. In a concurrent setting, that just means that it'll wait longer for some other process to let go of the lock, right?

Why does that happen at all (i.e. 5s seems like plenty of time)? Are many processes continuously writing to the cache? This mostly feels like a workaround, but I haven't debugged things, so not sure :\

pytools/persistent_dict.py
```diff
@@ -627,12 +628,24 @@ def store(self, key: K, value: V, _skip_if_present: bool = False) -> None:
...
-                                         "tried overwriting key")
+        if _skip_if_present:
+            self.conn.execute("INSERT OR IGNORE INTO dict VALUES (?, ?)",
+                              (keyhash, v))
+        else:
+            try:
+                self.conn.execute("INSERT INTO dict VALUES (?, ?)", (keyhash, v))
+            except sqlite3.IntegrityError as e:
+                if hasattr(e, "sqlite_errorcode"):
+                    if e.sqlite_errorcode == 1555:
```
alexfikl, line 639:
```suggestion
                    if e.sqlite_errorcode == sqlite3.SQLITE_CONSTRAINT_PRIMARYKEY:
```
? The constants are defined in on my system. Is that not the case in general?

matthiasdiener: Good catch, thanks!

pytools/persistent_dict.py
matthiasdiener: > If I understand correctly, the default timeout is 5s and this increases it to 60s. In a concurrent setting, that just means that it'll wait longer for some other process to let go of the lock, right?

Yes, this is my understanding as well.

> Why does that happen at all (i.e. 5s seems like plenty of time)? Are many processes continuously writing to the cache? This mostly feels like a workaround, but I haven't debugged things, so not sure :\

We aren't sure how this dictionary will be used downstream, it could be that thousands of processes are hitting the same dict at the same time. This change restores the timeout from the previous implementation: https://github.com/inducer/pytools/blob/f0846691d9b2c3575583b93570cff5fe237a794c/pytools/persistent_dict.py#L135-L136

alexfikl: > We aren't sure how this dictionary will be used downstream, 

That's fair, but we mostly know how `pyopencl` uses it. Why does it fail there in the tests? My understanding is that the tests run with `pytest-xdist` with `-n 4`, so it shouldn't have that many concurrent accesses. Does it?

> it could be that thousands of processes are hitting the same dict at the same time. This change restores the timeout from the previous implementation: 

Hmm, I don't think that sqlite is meant to be that concurrent in writes. Will that work reasonably well?

matthiasdiener: > > We aren't sure how this dictionary will be used downstream,
> 
> That's fair, but we mostly know how `pyopencl` uses it. Why does it fail there in the tests? My understanding is that the tests run with `pytest-xdist` with `-n 4`, so it shouldn't have that many concurrent accesses. Does it?

We are still trying to figure out the reason for the slowness observed in #227. So far, this seems to only affect the beers + @inducer's laptop, which is why I didn't see this issue earlier (I had tested on my Macs, Lassen, as well as GitHub CI, which are all 2 orders of magnitude faster than the beers in these tests).

> > it could be that thousands of processes are hitting the same dict at the same time. This change restores the timeout from the previous implementation:
> 
> Hmm, I don't think that sqlite is meant to be that concurrent in writes. Will that work reasonably well?

It can be a matter of the right configuration, but people have reported thousands of reads/writes per second with sqlite (see e.g. https://www.reddit.com/r/golang/comments/16xswxd/comment/k34ppfo/)

PR #229, opened 2024-06-04
PersistentDict: allow setting sync

matthiasdiener:
Should address the performance issues seen in #227.


inducer:
Thanks! Tweaked the warning and made some thing kw-only. LGTM!

PR #230, opened 2024-06-04
PersistentDict: smarter WOPD.store

matthiasdiener:
Passes through IntegrityErrors that are not ReadOnlyEntryErrors


inducer:
Thanks!

Issue #233, opened 2024-06-09
PersistentDict: multithread support?


matthiasdiener:
See e.g. https://gitlab.tiker.net/inducer/pyopencl/-/jobs/703834:

```
Exception ignored in: <function _PersistentDictBase.__del__ at 0x7fe3a0ff7ce0>
Traceback (most recent call last):
  File "/builds/inducer/pyopencl/boxtree/.conda-root/envs/testing/lib/python3.11/site-packages/pytools/persistent_dict.py", line 506, in __del__
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140615692060352 and this is thread id 140615717181248.
```

To reproduce in boxtree:

```console
$ pytest -n 4 test/test_tools.py
```

inducer:
Thanks for spotting that! I had seen it fly by, but then lost track of the CI run where it happened. I'm not quite sure either why it's happening (there should only be a single thread?), or what to do about it. What's your take?

inducer:
Also: this doesn't seem severe enough (yet?) to merit a yank, at least until we understand it better.

vincefn:
Hi @matthiasdiener @inducer - I just hit this bug yesterday using pyopencl in a multi-threaded environment (old code, now I'm more relying on multiprocessing but the code still works so...).

What is strange is that it [fails in one of my CI pipelines](https://gitlab.esrf.fr/favre/PyNX/-/jobs/1099410), but when I run the same testsuite from my account with the same environment setup (and same computer), it passes. I don't know how the type of account can affect the caching mechanism ?

I [just confirmed](https://gitlab.esrf.fr/favre/PyNX/-/jobs/1101359) that if I revert to pytools 2024.1.3 and pyopencl 2024.2.5, this CI failure disappears in my tests...

vincefn:
Followup - I just test this using a terminal in a remote desktop (so, _not a login shell_) and I could reproduce the issue. So that may explaine the difference between a CI job and a test in a login shell. Nasty.

matthiasdiener:
Thanks for the infos @inducer @vincefn. I'm currently working on a solution similar to https://ricardoanderegg.com/posts/python-sqlite-thread-safety/, since most sqlite builds should be threadsafe already, it is the Python layer that does the thread-related extra check that fails.

EdgarGF93:
Just to note that we have this issue with the package pyFAI while using pytools 2024.1.5, but it works with 2024.1.13

[...]

PR #234, opened 2024-06-10
PersistentDict: thread safety

matthiasdiener:
Workaround for #233.

cc: https://github.com/illinois-ceesd/logpyle/issues/106 https://github.com/illinois-ceesd/logpyle/issues/107 https://github.com/illinois-ceesd/logpyle/issues/108

Based partly on https://ricardoanderegg.com/posts/python-sqlite-thread-safety/.

Notes:
- Full thread safety ("serialized mode" in SQLite) is the default mode of SQLite (https://sqlite.org/threadsafe.html)
- There appears to be no way thread safety can be set from the Python interface

The only other alternative I can see is to use a fully multithreaded implementation such as https://github.com/piskvorky/sqlitedict


pytools/__init__.py
```diff
@@ -196,6 +196,12 @@
 
 .. autofunction:: unique
 
+SQLite-related functions
+------------------------
+
+.. autofunction:: get_sqlite3_thread_safety_level
+.. autofunction:: is_sqlite3_fully_threadsafe
```
inducer, line 203: I'm not sure I'd like to document these, since it encodes things that feel like implementation details. Maybe push this off into `pytools._sqlite`?

pytools/persistent_dict.py
```diff
@@ -459,9 +459,13 @@ def __init__(self, identifier: str,
         self.container_dir = container_dir
         self._make_container_dir()
 
+        from pytools import is_sqlite3_fully_threadsafe
+
         # isolation_level=None: enable autocommit mode
         # https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions
-        self.conn = sqlite3.connect(self.filename, isolation_level=None)
+        self.conn = sqlite3.connect(self.filename,
```
inducer, line 466: Maybe error (or warn?) in the unsafe case?

pytools/__init__.py
```diff
@@ -2975,6 +2981,51 @@ def unique(seq: Sequence[T]) -> Iterator[T]:
...
+        if sys.version_info < (3, 11):
+            assert threadsafety_value == 1
+        else:
+            assert threadsafety_value_db == sqlite3.threadsafety
+
+    return threadsafety_value_db
+
+
+def is_sqlite3_fully_threadsafe() -> bool:
+    """Check if the underlying SQLite library is fully thread-safe."""
```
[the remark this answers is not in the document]
matthiasdiener: I think our thoughts crossed here ;-) [ab252a6](https://github.com/inducer/pytools/pull/234/commits/ab252a6acae2f409835cd4f34b30cd1fe83746dd)

pytools/persistent_dict.py
matthiasdiener: Added a warning in f2f493e

pytools/__init__.py
matthiasdiener: Moving these functions into a separate module makes sense I think, but I thought these functions are pretty generic? Which implementation details do they encode?

matthiasdiener:
Another idea: in case the sqlite library is not built with thread safety, can we just use a mutex every time we use the connection (e.g., in `_exec_sql_fn`), and also set `check_same_thread=False`?

pytools/persistent_dict.py
matthiasdiener: Not necessary anymore, the mutex is always used.

matthiasdiener:
It seems it is crashing in the multithreaded case where the connection is shared among threads, in a configuration which should not crash even without the mutex :-(

```
  File "/home/runner/work/pytools/pytools/pytools/persistent_dict.py", line 778, in fetch
    stored_key, value = pickle.loads(row[0])
                        ^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'NoneType'
```

pytools/persistent_dict.py
inducer: > Not necessary anymore, the mutex is always used.

Could you explain what you mean here? I'm not sure I understand.

matthiasdiener: I meant that in https://github.com/inducer/pytools/pull/234/commits/4ca16cce520ec36497b3616718d4518f30851f3f, I removed the distinction between safe and unsafe cases, and ust using the mutex always.

pytools/__init__.py
matthiasdiener: > How cpython and threads and sqlite interact. The details of how this works at the moment is kind of insane ATM

I see - is this in reference to the translation between thread support levels in https://github.com/inducer/pytools/pull/234/commits/07412b2b08f63e7c6da8af4bc2424aa5856d0d10#diff-2d67e0c2f525bd7f05e4909f1248231d4b1f3b92d2f1be94487d5b2d4cb46b78R2992-R2996, or are there other problems in interacting between threads/sqlite? If so, could those be the issue for the failures I mentioned in https://github.com/inducer/pytools/pull/234#issuecomment-2159452812?

matthiasdiener:
Well, that turned out to be more ... interesting... than expected. The final error only happened with Python 3.12, which made debugging this more difficult.

This is ready for testing/review. 

@vincefn: Could you test this PR against the failure you had mentioned in #233? 

It does avoid the warning I had seen in the boxtree test.

vincefn:
Sorry it took a while - I could only reproduce that inside a gitlab CI test... But the tests which failed are passing with this fix.

PR #235, opened 2024-06-24
mypy: remove --follow-imports=skip checks

alexfikl:
This "fixes" the mypy failure on the CI:
https://github.com/inducer/pytools/actions/runs/9630439571/job/26561001591

As far as I can tell, that nice crash was caused by 
```
mypy --strict --follow-imports=skip pytools/persistent_dict.py
```
and seems to be related to https://github.com/python/mypy/issues/17396. I couldn't find any nicer workaround for it though..

cc @matthiasdiener

### DIFF
--- a/pytools/test/test_pytools.py
+++ b/pytools/test/test_pytools.py
@@ -172,7 +172,7 @@ def test_memoize_keyfunc():
     assert count[0] == 2
 
 
-def test_memoize_frozen():
+def test_memoize_frozen() -> None:
     from dataclasses import dataclass
 
     from pytools import memoize_method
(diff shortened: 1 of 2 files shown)


inducer:
Thanks for finding the mypy bug! I was puzzling over how to deal with that. I've cherry-picked one commit and done an alternate fix for the mypy issue in #237. Closing here.

alexfikl:
> Hm... TIL mypy behaves differently in CI and on my machine? Weird. The "fix" in #237 clearly isn't.

mypy caches quite heavily.. Maybe that's why? :\

inducer:
> mypy caches quite heavily.. Maybe that's why? :\ i.e. a `rm -rf .mypy_cache && ./run-mypy.sh` might help..

Tried that! It did not. Super odd.

alexfikl:
> Tried that! It did not. Super odd.

No idea then! For what it's worth, just tried #237 locally and it fails for me like on the CI.

PR #236, opened 2024-06-24
Unique improvements

matthiasdiener:
Better type annotation, test, doc.


matthiasdiener:
(mypy failure is fixed by #235)

inducer:
Thanks! (Partially) cherry-picked into #237 so that we only have to wait for CI once.

PR #237, opened 2024-06-24
Document convergence functionality

inducer:


matthiasdiener:
This also seems to work as a mypy workaround: https://github.com/inducer/pytools/pull/236/commits/4f9f5a63634d2e1ccbba3de53c4b9ef478d98e4b

alexfikl:
> This also seems to work as a mypy workaround: [4f9f5a6](https://github.com/inducer/pytools/commit/4f9f5a63634d2e1ccbba3de53c4b9ef478d98e4b) (from #236)

Haven't checked locally, but that should work!

Just removing `--follow-imports` goes through the existing imports and fails because they're not typed for `--strict`. That's why I also removed `--strict` in #235.

PR #238, opened 2024-06-27
Fix some typing issues

alexfikl:
While playing with #235 and #237 to fix the mypy issues, I had some more random fixes sitting around.

This essentially removes the `ignore_errors = True` in `setup.cfg` for some modules and fixes the errors that popped up. There's a bunch of stuff in `persistent_dict.py` that probably requires a careful look.

cc @matthiasdiener since you typed most of `PersistentDict`.

### DIFF
--- a/pytools/mpiwrap.py
+++ b/pytools/mpiwrap.py
@@ -13,7 +13,8 @@ import pytools.prefork  # pylint:disable=wrong-import-position
 pytools.prefork.enable_prefork()
 
 
-if Is_initialized():  # noqa pylint:disable=undefined-variable
+# pylint: disable-next=undefined-variable
+if Is_initialized():    # type: ignore[name-defined,unused-ignore] # noqa
     raise RuntimeError("MPI already initialized before MPI wrapper import")
 
 
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -2768,7 +2768,9 @@ def resolve_name(name):
 
 # {{{ unordered_hash
 
-def unordered_hash(hash_instance, iterable, hash_constructor=None):
+def unordered_hash(hash_instance: Any,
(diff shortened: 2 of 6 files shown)


pytools/mpiwrap.py
```diff
@@ -13,7 +13,8 @@
 pytools.prefork.enable_prefork()
 
 
-if Is_initialized():  # noqa pylint:disable=undefined-variable
+# pylint:disable=undefined-variable
```
inducer, line 16: I *think* this disables that warning for the whole file. Is that the intention?

inducer (commented):
Found one small wrinkle. Otherwise LGTM.

pytools/mpiwrap.py
alexfikl: My reading of their documentation
https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html#block-disables
is that if there's a statement on the next line, it disables it for that, and if there's not it disables it for the whole block.

I can put it back on the same line, it just looked a bit silly with 3 types of ignores :(

pytools/mpiwrap.py
```diff
@@ -13,7 +13,8 @@
...
+if Is_initialized():    # type: ignore[name-defined,unused-ignore] # noqa
```
alexfikl, line 17: I'm not getting the `named-defined` error locally, but it shows up on the CI, so I added the `unused-ignore`, which is not great :(

pytools/mpiwrap.py
inducer: Did you mean `disable-next`?

alexfikl: Yep :facepalm: Somehow I assumed that it just worked™ with the normal disable..

PR #240, opened 2024-07-03
pyproject: hack around to make editable install work

alexfikl:
Same issue as for `mypy` in https://github.com/inducer/modepy/pull/88.

This adds a few more commits with extra rules and fixes, so probably best looked at commit-by-commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,10 +10,10 @@ version = "2024.1.6"
 description = "A collection of tools for Python"
 readme = "README.rst"
 license = { text = "MIT" }
-requires-python = "~=3.8"
 authors = [
     { name = "Andreas Kloeckner", email = "inform@tiker.net" },
 ]
+requires-python = ">=3.8"
 classifiers = [
     "Development Status :: 4 - Beta",
     "Intended Audience :: Developers",
@@ -22,7 +22,7 @@ classifiers = [
     "License :: OSI Approved :: MIT License",
     "Natural Language :: English",
     "Programming Language :: Python",
-    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
     "Topic :: Scientific/Engineering",
     "Topic :: Scientific/Engineering :: Information Analysis",
     "Topic :: Scientific/Engineering :: Mathematics",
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -31,24 +36,32 @@ classifiers = [
...
 
 [project.urls]
-Homepage = "https://github.com/inducer/pytools/"
 Documentation = "https://documen.tician.de/pytools/"
+Homepage = "https://github.com/inducer/pytools/"
+
+[tool.setuptools.packages.find]
+include = [
+    "pytools*",
+]
```
alexfikl, line 60: I also had to add this because otherwise the wheel would contain some other folders as well (e.g. doc).

alexfikl: Hm, or not, I recall that being an issue with `modepy` or something.. Oh well..

alexfikl:
Hm, absolutely no idea why that one test failed on `pypy` :\
```
TypeError: expected some sort of stmt, but got '@py_assert3'
```
Makes it sound like it didn't like some pytest rewriting, but that's not new..

inducer:
:shrug: Let's just disable pypy testing then.

pyproject.toml
```diff
@@ -22,7 +22,12 @@ classifiers = [
     "License :: OSI Approved :: MIT License",
     "Natural Language :: English",
     "Programming Language :: Python",
-    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
inducer, line 30:
```suggestion
```
Life's too short to keep this up to date. Plus it's redundant with `requires-python`.

alexfikl: Yeah, agreed. I used https://github.com/tox-dev/pyproject-fmt to format this a bit and it seems to add those by default :\

alexfikl:
> 🤷 Let's just disable pypy testing then.

Any chance this is just on this branch?

inducer:
> Any chance this is just on this branch?

Don't care. Life's too short. :)

pyproject.toml
alexfikl: Removed in https://github.com/inducer/pytools/compare/1d13af56d6045ff02b087795cb4963f5c4c04623..a46be3f13216dd9781758fb96e91b0a3917e1154

alexfikl:
> Don't care. Life's too short. :)

Rerunning seems to have fixed that, so yey! This should be good to go now :grin:

PR #242, opened 2024-07-03
Use siphash for persistent hashing

inducer:
I don't know how much of a bottleneck the hash computation is in mirgecom compiles, but at the same time, I can't think of a good reason to not do this. Our use case very much falls into the "many small messages" category, so this graph is pretty compelling:

grafik

(from https://cr.yp.to/siphash/siphash-20120918.pdf)

A few things need to happen in order to make this go:
- [ ] ~~siphash24 would need to be [packaged for conda-forge](https://github.com/conda-forge/staged-recipes/pull/26856) (presumably straightforward)~~ (This can wait until next release and we package this for conda-forge.)
- [x] Someone would need to go through and change the hardcoded digests.
- [x] https://github.com/dnicolodi/python-siphash24/issues/6 (though we could hack around that)
- [x] https://github.com/inducer/loopy/pull/849

cc @matthiasdiener

### DIFF
--- a/.pylintrc-local.yml
+++ b/.pylintrc-local.yml
@@ -4,3 +4,4 @@
 - arg: ignored-modules
   val:
   - matplotlib
+  - siphash24
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ classifiers = [
 dependencies = [
     "platformdirs>=2.2",
     "typing-extensions>=4; python_version<'3.11'",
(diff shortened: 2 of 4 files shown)


inducer:
In addition, this will likely save considerable amounts of DRAM and disk.

inducer:
Python also uses [siphash internally](https://peps.python.org/pep-0456/) (for `hash`), but AFAIK that implementation is not individually accessible.

PR #243, opened 2024-07-06
Remove a bunch of unused pylint supressions

alexfikl:
This removes all the unused suppressions reported by `--enable=I0021`. I'm using pylint 3.2.5 on python 3.12.4, so I imagine not all of these will still pass on the CI, but hopefully some :grin: 

EDIT: That's weird.. I really didn't expect all those `no-member` ones to pass. Oh well!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
         runs-on: ${{ matrix.os }}
         strategy:
             matrix:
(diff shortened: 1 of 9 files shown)


alexfikl:
So the pypy CI failed again in https://github.com/inducer/pytools/actions/runs/9817472729/job/27108879263?pr=243
It seems to be a known issue https://github.com/pytest-dev/pytest/issues/11771

The issue seems to be fixed with pypy3.9, so this bumps the version for now..

PR #246, opened 2024-07-17
Revive Taggable constructor, in deprecated form

inducer:


pytools/tag.py
```diff
@@ -255,6 +254,15 @@ class Taggable:
     .. versionadded:: 2021.1
     """
 
+    if not TYPE_CHECKING:
+        def __init__(self, tags: FrozenSet[Tag] = frozenset()):
+            warn("The Taggable constructor is deprecated. "
+                 "Subclasses must declare their onwn storage for .tags. "
```
matthiasdiener, line 260:
```suggestion
                 "Subclasses must declare their own storage for .tags. "
```

inducer: Oof. This automerged before its time. :facepalm: Thanks for pointing this out.

PR #250, opened 2024-07-23
persistent_dict: make siphash24 optional

ahesford:
From #242:

> I don't know how much of a bottleneck the hash computation is in mirgecom compiles, but at the same time, I can't think of a good reason to not do this.

I maintain `pytools` and `pyopencl` on Void Linux and, as long as the hash interface is equivalent, making `siphash24` optional will slightly reduce the maintenance burden of carrying these packages alongside the rest of the numerical stack.

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -50,9 +50,15 @@ from typing import (
     TypeVar,
     cast,
 )
+from warnings import warn
 
-from siphash24 import siphash13
 
+try:
+    from siphash24 import siphash13 as _default_hash
+except ImportError:
+    warn("pytools.persistent_dict: unable to import 'siphash24.siphash13', "
+         "falling back to hashlib.sha256", stacklevel=1)
+    from hashlib import sha256 as _default_hash
 
 if TYPE_CHECKING:
     from _typeshed import ReadableBuffer
@@ -161,7 +167,7 @@ class KeyBuilder:
 
     # this exists so that we can (conceivably) switch algorithms at some point
     # down the road
-    new_hash: Callable[..., Hash] = siphash13
+    new_hash: Callable[..., Hash] = _default_hash
 
     def rec(self, key_hash: Hash, key: Any) -> Hash:
         """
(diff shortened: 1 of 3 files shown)


inducer:
Thanks for the contribution. I can see where you're coming from. At the same time, I'm a bit torn. Suppose a user fills a `PersistentDict`, then (say as a dependency of another package) installs siphash24. Suddenly the content of their `PersistentDict` became inaccessible. OTOH, they're used as caches most of the time, so who cares? :shrug: I don't know.

pytools/persistent_dict.py
```diff
@@ -51,8 +51,10 @@
     cast,
 )
 
-from siphash24 import siphash13
-
+try:
+    from siphash24 import siphash13 as _DEFAULT_HASH
+except ImportError
+    from hashlib import sha256 as _DEFAULT_HASH
```
inducer, line 57: Could you add a warning in this case that the recommended hash was not found?

inducer:
Could you look into the CI failures?

pytools/persistent_dict.py
ahesford: Done

ahesford:
CI failures were a mindless typo. I don't have a strong defense for this change, so if you think the value of enforcing consistency even as the package graph changes, I'll most likely just package `siphash24` for Void and be done with it.

inducer:
I'd be willing to give this a shot. But be warned: If it blows up on me in some way, I'll probably just revert it. Also, Ruff had some more complaints. Could you take a look?

ahesford:
That seems reasonable. The latest patch should pass all checks.

inducer:
https://pypi.org/project/pytools/2024.1.11/

ahesford:
Thanks!

matthiasdiener:
I don't know about this- now users of `PersistentDict` won't get the benefits of `siphash24` unless they install `siphash24` manually (but instead get a warning that does not clearly describe how to resolve the issue)...

This affects downstream packages too (see e.g. https://github.com/matthiasdiener/skvlite/actions/runs/10066243771/job/27827181383?pr=13)

inducer:
To be fair, the size of that SQLite file could be considered an implementation detail. 523ff36 rewords the warning so as to be more actionable.

PR #251, opened 2024-07-23
PersistentDict: make siphash24 optional in the tests

matthiasdiener:
Otherwise, those tests fail when siphash24 is not installed.

Followup of #250.

(Imho, no need for a new release just for this PR)


inducer:
Honestly, I'm comfortable with the tests requiring siphash24.

matthiasdiener:
> Honestly, I'm comfortable with the tests requiring siphash24.

Ok, that's also fine.

PR #252, opened 2024-07-23
tag.py: try importing from typing first

matthiasdiener:
_Please squash_

### DIFF
--- a/pytools/tag.py
+++ b/pytools/tag.py
@@ -39,7 +39,11 @@ from typing import (
 )
 from warnings import warn
 
-from typing_extensions import Self, dataclass_transform
+
+try:
+    from typing import Self, dataclass_transform
+except ImportError:
+    from typing_extensions import Self, dataclass_transform
 
 from pytools import memoize, memoize_method
 


matthiasdiener:
Just a heads-up @inducer, Python 3.13 may need `typing-extensions` too.

inducer:
`TypeAlias` was in 3.10? https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeAlias

matthiasdiener:
> `TypeAlias` was in 3.10? https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeAlias

It seems to be deprecated already, and perhaps removed in 3.13? https://docs.python.org/3/library/typing.html#typing.TypeAlias

inducer:
Yes, they introduced the `type` statement in 3.12. But that won't become usable for us until we go to a minimum version of 3.12, which will take another few years.

inducer:
(So `TypeAlias` will be around for that same amount of time.)

matthiasdiener:
What should we do to enable 3.13 compatibility? Require `typing_extensions` (like in this PR)?

alexfikl:
> What should we do to enable 3.13 compatibility? Require `typing_extensions` (like in this PR)?

I'm not sure I understand the issue. `TypeAlias` is not removed in python 3.13 and the docs say that removal is not planned:
```
Removal of [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is not currently planned, but users are encouraged to migrate to [type](https://docs.python.org/3/reference/simple_stmts.html#type) statements.
```
What errors are you seeing?

matthiasdiener:
The error [here](https://github.com/inducer/pytools/actions/runs/10067150871/job/27830067869):

```
________________ ERROR collecting pytools/test/test_pytools.py _________________
ImportError while importing test module '/home/runner/work/pytools/pytools/pytools/test/test_pytools.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.13.0-beta.4/x64/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
pytools/test/test_pytools.py:32: in <module>
    from pytools.tag import tag_dataclass
pytools/tag.py:42: in <module>
    from typing_extensions import Self, dataclass_transform
E   ModuleNotFoundError: No module named 'typing_extensions'
```

alexfikl:
> The error [here](https://github.com/inducer/pytools/actions/runs/10067150871/job/27830067869):

That looks like a little bug to me. If we're requiring `typing_extensions` conditionally in `pyproject.toml`, it should be
```python
try: 
	from typing import Self, dataclass_transform
except ImportError:
	from typing_extensions import Self, dataclass_transform
```

inducer:
@matthiasdiener Could you make a PR with the change suggested by @alexfikl? I agree it's a bug.

matthiasdiener:
The CI with 3.13 passed, so I just modified this PR (minus the additional CI changes), if that's ok.

alexfikl (approved):

inducer:
Thx!

PR #257, opened 2024-08-14
more unique functions

matthiasdiener:
_Please squash_

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -41,7 +41,6 @@ from typing import (
     Generic,
     Hashable,
     Iterable,
-    Iterator,
     List,
     Mapping,
     Optional,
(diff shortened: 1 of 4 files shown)


pytools/__init__.py
```diff
@@ -2997,6 +3003,49 @@ def unique(seq: Iterable[T]) -> Iterator[T]:
     :func:`itertools.groupby` (which removes consecutive duplicates)."""
     return iter(dict.fromkeys(seq))
 
+
+def unique_difference(*args: Iterable[T]) -> Iterator[T]:
```
inducer, line 3007: Would it make sense to have these return `Collection[T]` (and skip the `iter` call)?

matthiasdiener: Hmm, I'm not sure what to return instead of the `iter` - a list or tuple? We can't just return the dictionary I guess.

matthiasdiener: 9ece452 is a version of these functions returning lists. I don't know if that's better.

inducer: No, I thought you would just return the dictionary. A dictionary with key type `T` should count as a collection of type `T` I think.

inducer: I.e. no need to build a separate list.

matthiasdiener: What do you think of [7b352f3](https://github.com/inducer/pytools/pull/257/commits/7b352f3103f55b627cbf550d404b408638f05354)?

inducer: LGTM!

inducer:
Thx!

Issue #259, opened 2024-08-31
`keyb(1) == keyb(np.int64(1))` fails on ppc64 (big endian)


mgorny:
When running the test suite on ppc64 (big endian), I'm getting the following test failure:

```pytb
$ pytest pytools/test/test_persistent_dict.py::test_scalar_hashing
========================================================= test session starts =========================================================
platform linux -- Python 3.12.3, pytest-8.3.2, pluggy-1.5.0
rootdir: /home/mgorny/pytools
configfile: pyproject.toml
plugins: xdist-3.6.1
collected 1 item                                                                                                                      

pytools/test/test_persistent_dict.py F                                                                                          [100%]

============================================================== FAILURES ===============================================================
_________________________________________________________ test_scalar_hashing _________________________________________________________

    def test_scalar_hashing() -> None:
        keyb = KeyBuilder()
    
        assert keyb(1) == keyb(1)
        assert keyb(2) != keyb(1)
        assert keyb(1.1) == keyb(1.1)
        assert keyb(1+4j) == keyb(1+4j)
    
        try:
            import numpy as np
        except ImportError:
            return
    
        assert keyb(np.int8(1)) == keyb(np.int8(1))
        assert keyb(np.int16(1)) == keyb(np.int16(1))
        assert keyb(np.int32(1)) == keyb(np.int32(1))
        assert keyb(np.int32(2)) != keyb(np.int32(1))
        assert keyb(np.int64(1)) == keyb(np.int64(1))
>       assert keyb(1) == keyb(np.int64(1))
E       AssertionError: assert 'df3e66fafc316707' == 'f848714b281ce907'
E         
E         - f848714b281ce907
E         + df3e66fafc316707

pytools/test/test_persistent_dict.py:435: AssertionError
======================================================= short test summary info =======================================================
FAILED pytools/test/test_persistent_dict.py::test_scalar_hashing - AssertionError: assert 'df3e66fafc316707' == 'f848714b281ce907'
========================================================== 1 failed in 0.43s ==========================================================
```

I can reproduce on git 6c1bd3f42d02426ada42760116223f6aff57452b and 2024.1.13, with NumPy 2.1.0.

I'm guessing this is because `update_for_int()` forces little endian:

https://github.com/inducer/pytools/blob/6c1bd3f42d02426ada42760116223f6aff57452b/pytools/persistent_dict.py#L278-L286

but `update_for_numpy_scalar()` uses native byte order:

[...]

PR #260, opened 2024-08-31
KeyBuilder: use system byteorder to match numpy

inducer:
Closes gh-259

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -174,6 +174,12 @@ class KeyBuilder:
         may stop working as early as 2022.
 
         .. versionadded:: 2021.2
+
+    .. note::
+
+        Some key-building uses system byte order, so the built keys may not match
+        across different systems. It would be desirable to fix this, but this is
+        not yet done.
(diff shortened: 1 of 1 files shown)


pytools/persistent_dict.py
```diff
@@ -280,7 +286,10 @@ def update_for_int(key_hash: Hash, key: int) -> None:
         sz = 8
         while True:
             try:
-                key_hash.update(key.to_bytes(sz, byteorder="little", signed=True))
+                # Must match system byte order so that numpy and this
+                # generate the same string of bytes.
+                # https://github.com/inducer/pytools/issues/259
+                key_hash.update(key.to_bytes(sz, byteorder=sys.byteorder, signed=True))
```
matthiasdiener, line 292: Out of curiosity, do we actually need to match the numpy representation for ints? Could we just use something like `str(key)` here (similar to what is used for float and complex)?

inducer: I guess that'd be better. Could you do a PR?

PR #261, opened 2024-09-03
KeyBuilder: change int to direct str representation

matthiasdiener:
Followup of #260.

### DIFF
--- a/pytools/test/test_persistent_dict.py
+++ b/pytools/test/test_persistent_dict.py
@@ -439,6 +439,9 @@ def test_scalar_hashing() -> None:
     assert keyb(1.0) != keyb("1.0")  # '1.0' uses hex representation
     assert keyb(1+1j) == keyb("(1+1j)")
 
+    assert keyb(np.float64(1.0)) != keyb(1.0)
+    assert keyb(np.complex128(1.0+2.0j)) != keyb(1.0+2.0j)
+
     assert keyb(np.longlong(1)) == keyb(np.longlong(1))
 
     assert keyb(np.float16(1.1)) == keyb(np.float16(1.1))
--- a/pytools/test/test_persistent_dict.py
+++ b/pytools/test/test_persistent_dict.py
@@ -432,8 +432,12 @@ def test_scalar_hashing() -> None:
     assert keyb(np.int32(1)) == keyb(np.int32(1))
     assert keyb(np.int32(2)) != keyb(np.int32(1))
     assert keyb(np.int64(1)) == keyb(np.int64(1))
-    assert keyb(1) == keyb(np.int64(1))
+    assert keyb(1) != keyb(np.int64(1))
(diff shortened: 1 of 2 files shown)


pytools/test/test_persistent_dict.py
```diff
@@ -432,9 +432,13 @@ def test_scalar_hashing() -> None:
     assert keyb(np.int32(1)) == keyb(np.int32(1))
     assert keyb(np.int32(2)) != keyb(np.int32(1))
     assert keyb(np.int64(1)) == keyb(np.int64(1))
-    assert keyb(1) == keyb(np.int64(1))
+    assert keyb(1) != keyb(np.int64(1))
     assert keyb(1) != keyb(np.int32(1))
 
+    assert keyb(1) == keyb("1")
+    assert keyb(1.0) != keyb("1.0")  # '1.0' uses hex representation
+    assert keyb(1+1j) == keyb("(1+1j)")
```
matthiasdiener, line 440: Not sure if this is an issue. If it is, we could change the KeyBuilder to `(int, str(key)`

pytools/test/test_persistent_dict.py
```diff
@@ -432,9 +432,13 @@ def test_scalar_hashing() -> None:
     assert keyb(np.int32(1)) == keyb(np.int32(1))
     assert keyb(np.int32(2)) != keyb(np.int32(1))
     assert keyb(np.int64(1)) == keyb(np.int64(1))
-    assert keyb(1) == keyb(np.int64(1))
+    assert keyb(1) != keyb(np.int64(1))
```
inducer, line 435: Wait, I thought the goal was to keep them the same?

matthiasdiener: From your answer to [my question](https://github.com/inducer/pytools/pull/260#discussion_r1740209871) in #260, my interpretation was that they don't need to be the same. Changing the code like in this PR perhaps also addresses #262.

mgorny:
Just FYI, the tests pass on PPC64 BE with this patch applied.

pytools/test/test_persistent_dict.py
inducer: They need to be the same. The fundamental property of a hash function $h$ is that $x = y \Rightarrow h(x) = h(y)$. Since
```
>>> np.int32(1) == 1
True
```
our hands are tied. I'm sorry if my early comment was ambiguous.

inducer:

pytools/test/test_persistent_dict.py
matthiasdiener: Hmm, we already break this principle in some (many?) places:

```python
>>> from pytools.persistent_dict import KeyBuilder
>>> keyb = KeyBuilder()
>>> 1.0 == 1
True
>>> keyb(1.0) == keyb(1)
False
```

PR #265, opened 2024-10-17
Table: add raw output

matthiasdiener:
_Please squash_

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1756,6 +1756,38 @@ class Table:
 
         return "\n".join(lines)
 
+    def raw(self) -> str:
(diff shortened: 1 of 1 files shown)


pytools/__init__.py
```diff
@@ -1756,6 +1766,38 @@ def latex(self,
 
         return "\n".join(lines)
 
+    def raw(self) -> str:
```
inducer, line 1769:
```suggestion
    def text_without_markup(self) -> str:
```
? (I'm not liking `raw` very much as a name.)

inducer (commented):
LGTM, just quibbling about the name.

PR #266, opened 2024-10-23
Bump min Python to 3.10

alexfikl:
Had some time to waste and tried bumping this up to 3.10. `ruff` was very noisy, but easily fixable :D

There were a lot of places that wanted `zip(..., strict=True)` and will need to check if it actually makes sense (?).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
         runs-on: ${{ matrix.os }}
         strategy:
             matrix:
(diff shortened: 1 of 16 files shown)


alexfikl:
> This looks great! IMO, let's just do this. Any reason it's still draft?

I mostly threw it here, but thought it would wait for all the `pymbolic` changes to trickle through first. This also switched to a bunch of 3.10-only syntax, so it may break downstream CIs running on older versions?

inducer:
These are good points. I agree, it should probably wait.

PR #267, opened 2024-11-12
KeyBuilder: support np.bool{,_} for numpy<2

matthiasdiener:
The attached test fails with numpy 1.26.4 without this PR.

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -221,7 +221,7 @@ class KeyBuilder:
                             method = self.update_for_specific_dtype
 
                     # Hashing numpy scalars
-                    elif isinstance(key, np.number):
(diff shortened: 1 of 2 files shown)


alexfikl:
Maybe it's worth adding a CI with some oldest supported numpy?

inducer:
> Maybe it's worth adding a CI with some oldest supported numpy?

:shrug: In an ideal world maybe. But given that https://github.com/scipy/oldest-supported-numpy is deprecated, I don't want to maintain tables of which numpy is the oldest-supported, so I think we'll find these issues as they arise, and I think I'm vaguely OK with that.

inducer:
https://pypi.org/project/pytools/2024.1.15/

PR #270, opened 2024-11-27
mypy: add strict typing to prefork

alexfikl:
Mostly straightforward, but the socket communication is pretty much ignored for typing.

Don't quite recall, but I think some other package required typing `call_capture_output`, so went ahead and typed the whole thing.

### DIFF
--- a/pytools/prefork.py
+++ b/pytools/prefork.py
@@ -3,21 +3,65 @@ allows the forking of *one* helper child process before OpenMPI
 initialization that can do the forking for the fork-challenged
 parent process.
 
-Since none of this is MPI-specific, it got parked in pytools.
+Since none of this is MPI-specific, it got parked in :mod:`pytools`.
+
+.. autoexception:: ExecError
+    :show-inheritance:
+
+.. autoclass:: Forker
+.. autoclass:: DirectForker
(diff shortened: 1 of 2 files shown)


pytools/prefork.py
```diff
@@ -164,40 +210,55 @@ def _remote_invoke(self, name, *args, **kwargs):
...
+        return result
 
-    def call_capture_output(self, cmdline, cwd=None, error_on_nonzero=True):
-        return self._remote_invoke("call_capture_output", cmdline, cwd,
+    def call_capture_output(self,
+                            cmdline: Sequence[str],
+                            cwd: str | None = None,
+                            error_on_nonzero: bool = True,
+                            ) -> tuple[int, bytes, bytes]:
+        return self._remote_invoke("call_capture_output", cmdline, cwd,  # type: ignore[no-any-return,unused-ignore]
```
alexfikl, line 236: This is a bit annoying.. the non-strict mypy complains about the `unused-ignore` and the strict mypy complains about the `no-any-return` :(

inducer: You could change the `Any`s to `object`s maybe?

inducer:
Thanks!

PR #272, opened 2024-11-28
Opt frozen dataclass

inducer:


pytools/__init__.py
```diff
@@ -903,7 +910,7 @@ def new_inner(*args: P.args, **kwargs: P.kwargs) -> R:
         return new_inner
 
 
-class keyed_memoize_in(Generic[P, R]):  # noqa: N801
```
alexfikl, line 906: Why is this not needed? I thought it constrained the `__call__` inputs

inducer: The problem I encountered is that `R` does not get bound at construction time, and so mypy helpfully binds it to `Never`, which it expects as the return value of the function. pyright seemed to agree that `R` would not get bound.

`__call__` is still an `R`-generic function.

PR #273, opened 2024-12-02
`opt_frozen_dataclass`: Enable hashing with -O

inducer:
Inspired by https://github.com/inducer/pytato/pull/563.

cc @matthiasdiener (what do you think?)

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -1,5 +1,6 @@
 __copyright__ = """
 Copyright (C) 2009-2013 Andreas Kloeckner
+Copyright (C) 2013- University of Illinois Board of Trustees
 Copyright (C) 2020 Matt Wala
 """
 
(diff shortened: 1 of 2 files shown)


matthiasdiener:
Dunno, seems risky 😬

matthiasdiener:
Maybe replacing `unsafe_hash: bool = False,` with `unsafe_hash: bool = None,` could be an option?

inducer:
> Dunno, seems risky 😬

I'd argue it's needed for correctness. Otherwise, these classes are hashable when running without `-O` and not hashable when running with `-O`, which is inconsistent. They're supposed to be semantically frozen classes, as the name suggests. I'll add a comment as to why that's there.

inducer:
> Maybe replacing `unsafe_hash: bool = False,` with `unsafe_hash: bool = None,` could be an option?

Sure, I could see that.

inducer:
Paging @alexfikl for a third set of eyes: Is all this just a terrible idea and not worth it?

pytools/__init__.py
```diff
@@ -3000,16 +3000,38 @@ def opt_frozen_dataclass(
...
+        and vice versa. To ensure frozen-ness is applied predictably in all
+        scenarios (mainly :data:`__debug__` on and off), it is strongly recommended
+        that all dataclasses inheriting from ones with this decorator *also*
+        use this decorator. There are no run-time checks to make sure of this.
+
     .. versionadded:: 2024.1.18
     """
     def map_cls(cls: type[T]) -> type[T]:
+        # This ensures that the resulting dataclass is hashable with and without
+        # __debug__, unless the user overrides unsafe_hash.
```
matthiasdiener, line 3019:
```suggestion
        # __debug__, unless the user overrides unsafe_hash, or provides their own __hash__ method.
```

pytools/__init__.py
```diff
@@ -3000,16 +3000,38 @@ def opt_frozen_dataclass(
...
+        if (eq
+                and not __debug__
+                and unsafe_hash is None
+                and "__hash__" not in cls.__dict__):
+            loc_unsafe_hash = True
+        else:
+            loc_unsafe_hash = False
+
```
matthiasdiener, line 3027: I think this still doesn't allow the user to pass `unsafe_hash=True`?

pytools/__init__.py
```diff
@@ -3000,16 +3000,38 @@ def opt_frozen_dataclass(
...
     def map_cls(cls: type[T]) -> type[T]:
+        # This ensures that the resulting dataclass is hashable with and without
+        # __debug__, unless the user overrides unsafe_hash.
+        if (eq
+                and not __debug__
+                and unsafe_hash is None
+                and "__hash__" not in cls.__dict__):
+            loc_unsafe_hash = True
+        else:
+            loc_unsafe_hash = False
```
matthiasdiener, line 3026: maybe:

```suggestion
            loc_unsafe_hash = unsafe_hash
```

?

inducer: Addressed, thanks!

pytools/__init__.py
inducer: Addressed, thanks!

pytools/__init__.py
inducer: Addressed, thanks!

matthiasdiener:
Side note: [this table](https://github.com/python/cpython/blob/3.13/Lib/dataclasses.py#L121) is really helpful when thinking through this.

pytools/__init__.py
```diff
@@ -3000,16 +3000,41 @@ def opt_frozen_dataclass(
     this decorator avoid when the interpreter runs with "optimization"
```
matthiasdiener, line 3000:
```suggestion
    this decorator avoids when the interpreter runs with "optimization"
```

pytools/__init__.py
```diff
@@ -3000,16 +3000,41 @@ def opt_frozen_dataclass(
...
     enabled.
 
+    The resulting dataclass supports hashing unless *eq* is set to *False*,
+    if *unsafe_hash* is left at the default or set to *True*.
```
matthiasdiener, line 3004:
```suggestion
    The resulting dataclass supports hashing, even when it is not actually frozen,
    unless *eq* is set to *False*, if *unsafe_hash* is left at the default or set to *True*.
```

matthiasdiener (approved):
Besides the minor wording comments, this LGTM. Thanks!

alexfikl:
> Paging @alexfikl for a third set of eyes: Is all this just a terrible idea and not worth it?

Hm, I haven't followed this much, but if I understand correctly.. we want to turn off `frozen` when in `-O` mode because it's slower, but that also turns off hashing by default, right? And this is meant to sneakily add back hashing as if the class was frozen?

inducer:
Thanks all for your help in getting this into shape. I'll do another release as soon as this is in.

inducer:
https://pypi.org/project/pytools/2024.1.19/

PR #274, opened 2024-12-03
opt_frozen_dataclass: more improvements

matthiasdiener:
- raise an error when specifying `frozen`
- add tests
- add support for specifying `weakref_slot`

targets #273.

### DIFF
--- a/pytools/__init__.py
+++ b/pytools/__init__.py
@@ -2991,8 +2991,7 @@ def opt_frozen_dataclass(
             match_args: bool = True,
             kw_only: bool = False,
             slots: bool = False,
-            # Added in 3.11.
-            # weakref_slot: bool = False
+            **kwargs: Any,  # Extra, version dependent arguments (weakref_slot in 3.11)
          ) -> Callable[[type[T]], type[T]]:
     """Like :func:`dataclasses.dataclass`, but marks the dataclass frozen
     only if :data:`__debug__` is active. Frozen dataclasses have a ~20%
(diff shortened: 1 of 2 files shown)


matthiasdiener:
What do you think @inducer?

pytools/__init__.py
```diff
@@ -3014,13 +3013,22 @@ def opt_frozen_dataclass(
...
+        raise ValueError("frozen must not be specified in opt_frozen_dataclass")
+
     def map_cls(cls: type[T]) -> type[T]:
         # This ensures that the resulting dataclass is hashable with and without
         # __debug__, unless the user overrides unsafe_hash or provides their own
         # __hash__ method.
+
+        # Make it possible to override 'frozen' in the class definition for testing.
+        # It would be nice to have something like https://discuss.python.org/t/allow-debug-to-be-set-at-runtime/64840
+        loc_frozen = __debug__ if "_frozen_override" not in cls.__dict__ else False
```
inducer, line 3027: Could you explain the use case?

inducer, line 3027: `frozen_override` could be in `kwargs` maybe?

pytools/test/test_dataclasses.py
inducer, line 1: I was thinking this needed tests. Thanks for adding them!

pytools/__init__.py
```diff
@@ -2991,8 +2991,7 @@ def opt_frozen_dataclass(
             match_args: bool = True,
             kw_only: bool = False,
             slots: bool = False,
-            # Added in 3.11.
-            # weakref_slot: bool = False
+            **kwargs: Any,  # Extra, version dependent arguments (weakref_slot in 3.11)
```
inducer, line 2994: Is a `kwargs`-only `ParamSpec` a thing?

inducer:
Thanks! I've rolled  most of this into #273, with attribution. We can tackle the override later on if it's still needed.

pytools/__init__.py
matthiasdiener: It's just meant for testing so that we can "fake" setting `__debug__=False`.

matthiasdiener: Yeah, I think that's a good idea.

PR #275, opened 2024-12-04
also run dataclasses test with optimizations enabled

matthiasdiener:
_Please squash_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,10 @@ jobs:
                 curl -L -O https://gitlab.tiker.net/inducer/ci-support/raw/main/build-and-test-py-project.sh
                 . ./build-and-test-py-project.sh
 
+                # Also run with optimizations turned on, since opt_frozen_dataclass
+                # depends on the __debug__ setting.
+                python -O -m pytest pytools/test/test_dataclasses.py
+
     pytest_nonumpy:
         name: Pytest without Numpy
         runs-on: ubuntu-latest
--- a/pytools/test/test_dataclasses.py
+++ b/pytools/test/test_dataclasses.py
@@ -44,10 +44,13 @@ def test_opt_frozen_dataclass() -> None:
     assert a == A(1)
 
     # Needs to be frozen by default
-    with pytest.raises(AttributeError):
+    if __debug__:
+        with pytest.raises(AttributeError):
+            a.x = 2  # type: ignore[misc]
+    else:
         a.x = 2  # type: ignore[misc]
 
-    assert a.__dataclass_params__.frozen is True  # type: ignore[attr-defined]  # pylint: disable=no-member
+    assert a.__dataclass_params__.frozen is __debug__  # type: ignore[attr-defined]  # pylint: disable=no-member
 
     # }}}
 


matthiasdiener:
With these changes, we perhaps won't need a way to provide a `frozen_override` option like in #274.

matthiasdiener:
By the way, the Py3.x CI is currently testing 3.12, should we change this to Py3.13?

alexfikl:
> By the way, the `Py3.x` CI is currently testing 3.12 (which thus gets tested twice); should we change this to `Py3.13`?

That's weird. I thought Python 3.13 was available on Github?
https://github.com/actions/python-versions/blob/main/versions-manifest.json

matthiasdiener:
> > By the way, the `Py3.x` CI is currently testing 3.12 (which thus gets tested twice); should we change this to `Py3.13`?
> 
> That's weird. I thought Python 3.13 was available on Github. Any idea why it's not picked up? https://github.com/actions/python-versions/blob/main/versions-manifest.json

Yeah, it is available. But for some reasons `Py3.x` defaults to 3.12 on Ubuntu (MacOS gets 3.13 already 🤷 ).

inducer:
> By the way, the `Py3.x` CI is currently testing 3.12 (which thus gets tested twice); should we change this to `Py3.13`?

Nah, they'll update it soon enough. Don't have time to track Github's Python version shenanigans in detail.

.github/workflows/ci.yml
```diff
@@ -97,6 +97,10 @@ jobs:
                 curl -L -O https://gitlab.tiker.net/inducer/ci-support/raw/main/build-and-test-py-project.sh
                 . ./build-and-test-py-project.sh
 
+                # Also run with optimizations turned on, since opt_frozen_dataclass
+                # depends on the __debug__ setting.
+                python -O -m pytest pytools/test/test_dataclasses.py
```
alexfikl, line 102: Unrelated, but how does this work? I thought `-O` disabled assertions, but `pytest` does some assertion hackery to make the tests fail and whatnot, so combining them seems very odd.

inducer: Ah! Great question. I had not thought about this. It *used* to be the case that pytest got upset when you ran it with `python -O`, but that's (clearly!) no longer the case. So something else must be happening. A litte bit of digging turns up that pytest rewrites the AST to swap out `assert` statements for something [entirely different](https://pybites.blogspot.com/2011/07/behind-scenes-of-pytests-new-assertion.html), which, IIUC, makes it OK for pytest to run with `python -O`.

matthiasdiener: I'm not sure _how_ pytest re-enables assertions when running with -O, but I think they rewrite them to provide more information? (see e.g. https://github.com/pytest-dev/pytest/blob/9d4f36d87dae9a968fb527e2cb87e8a507b0beb3/src/_pytest/assertion/rewrite.py)

They seem to work fine as long as they occur in the test module. See this warning from CI:

```
=============================== warnings summary ===============================
.env/lib/python3.12/site-packages/_pytest/config/__init__.py:1278
  /home/runner/work/pytools/pytools/.env/lib/python3.12/site-packages/_pytest/config/__init__.py:1278: PytestConfigWarning: assertions not in test modules or plugins will be ignored because assert statements are not executed by the underlying Python interpreter (are you using python -O?)
  
    self._warn_about_missing_assertion(mode)
```

We set -O not because of assertions, but to modify the value of `__debug__`, which is what `opt_frozen_dataclass` looks at.

alexfikl: Very cool! Thanks! I'll read those more carefully to see what they're doing there :grin:

PR #278, opened 2024-12-11
Allow `get_reachable_nodes` to be restricted to partitions

a-alveyblanc:
Updates `get_reachable_nodes` to take an optional set containing nodes that should be ignored (effectively partition points) in an undirected graph. Also updates the corresponding test to account for this new feature.

### DIFF
--- a/pytools/graph.py
+++ b/pytools/graph.py
@@ -83,6 +83,7 @@ from dataclasses import dataclass
 from typing import (
     Any,
     Generic,
+    Optional,
     Protocol,
     TypeAlias,
     TypeVar,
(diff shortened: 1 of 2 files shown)


pytools/graph.py
```diff
@@ -566,24 +567,39 @@ def undirected_graph_from_edges(
 
 def get_reachable_nodes(
         undirected_graph: GraphT[NodeT],
-        source_node: NodeT) -> frozenset[NodeT]:
+        source_node: NodeT,
+        exclude_nodes: Optional[set[NodeT]] = None) -> frozenset[NodeT]:
```
inducer, line 571:
```suggestion
        exclude_nodes: Optional[Collection[NodeT]] = None) -> frozenset[NodeT]:
```

pytools/graph.py
```diff
@@ -566,24 +567,39 @@ def undirected_graph_from_edges(
...
     """
     Returns a :class:`frozenset` of all nodes in *undirected_graph* that are
-    reachable from *source_node*.
+    reachable from *source_node*. Reachability of nodes from *source_node* can
+    be restricted by supplying a set of nodes to be excluded from consideration.
+    Effectively, this amounts to traversing partitions of *undirected_graph*
+    with partition points defined by *exclude_nodes*.
```
inducer, line 577: Rephrase this in terms of nodes and edges. I find the "partition" terminology to be unclear.

pytools/graph.py
```diff
@@ -566,24 +567,39 @@ def undirected_graph_from_edges(
...
     while nodes_to_visit:
         current_node = nodes_to_visit.pop()
         nodes_visited.add(current_node)
 
-        neighbors = undirected_graph[current_node]
-        nodes_to_visit.update({node
-                               for node in neighbors
-                               if node not in nodes_visited})
+        if current_node in exclude_nodes:
+            continue
```
inducer, line 591: This seems redundant with the check below, except for the case of the initial node, which can be handled above.

pytools/graph.py
inducer, line 577: Document what happens if the source node is excluded.

pytools/graph.py
```diff
@@ -566,24 +567,39 @@ def undirected_graph_from_edges(
...
+
+        new_nodes_to_visit = set()
+        for node in undirected_graph[current_node]:
+            if node not in nodes_visited and node not in exclude_nodes:
+                new_nodes_to_visit.add(node)
+
+        nodes_to_visit.update(new_nodes_to_visit)
 
-    return frozenset(nodes_visited)
+    return frozenset(reachable_nodes)
```
inducer, line 602: Recast this as a generator expression?

inducer (commented):
Thanks! Some comments below.

inducer:
Btw: I rebased out the fixes for RUF052. [I campaigned](https://github.com/astral-sh/ruff/issues/14796) to get function arguments removed from that rule, and a new ruff with that change got released today--so we were just waiting that out.

inducer:

a-alveyblanc:
Ready for another look, @inducer

inducer:
I'll roll a release once #279 is in.

inducer:
https://pypi.org/project/pytools/2024.1.21/

PR #280, opened 2024-12-29
Type gnitsam functions

alexfikl:


pytools/__init__.py
```diff
@@ -1438,23 +1453,42 @@ def generate_nonnegative_integer_tuples_summing_to_at_most(n, length):
...
 
-def generate_all_integer_tuples_below(n, length, least_abs=0):
+def generate_all_integer_tuples_below(
+        n: int, length: int, least_abs: int = 0
+    ) -> Iterator[tuple[int, ...]]:
     return _pos_and_neg_adaptor(generate_nonnegative_integer_tuples_below(
         n, length, least_abs))
 
 
+@overload
```
inducer, line 1479: Are all these overloads really needed? (I.e. wouldn't `Sequence[T]` or `Iterable[T]` automatically do the right thing for `str`?)

pytools/__init__.py
```diff
@@ -1438,23 +1453,42 @@ def generate_nonnegative_integer_tuples_summing_to_at_most(n, length):
...
+def generate_permutations(original: tuple[T, ...]) -> Iterator[tuple[T, ...]]: ...
+
+@overload
+def generate_permutations(original: str) -> Iterator[str]: ...
+
+@overload
+def generate_permutations(original: bytes) -> Iterator[bytes]: ...
+
+
 def generate_permutations(original):
```
inducer, line 1492: Shouldn't this also be typed?

pytools/__init__.py
```diff
@@ -1465,12 +1499,25 @@ def generate_permutations(original):
     else:
         for perm_ in generate_permutations(original[1:]):
             for i in range(len(perm_)+1):
-                # nb str[0:1] works in both string and list contexts
+                # NOTE: ary[0:1] works in both string and list contexts
                 yield perm_[:i] + original[0:1] + perm_[i:]
 
 
+@overload
```
inducer, line 1506: As above.

pytools/__init__.py
```diff
@@ -1465,12 +1499,25 @@ def generate_permutations(original):
...
+def generate_unique_permutations(original: tuple[T, ...]) -> Iterator[tuple[T, ...]]: ...  # noqa: E501
+
+@overload
+def generate_unique_permutations(original: str) -> Iterator[str]: ...
+
+@overload
+def generate_unique_permutations(original: bytes) -> Iterator[bytes]: ...
+
+
 def generate_unique_permutations(original):
```
inducer, line 1516: As above.

inducer (commented):
Thanks! Two-ish (?) questions below.

pytools/__init__.py
alexfikl: I'm not sure.. my understanding of the docs https://docs.python.org/3/library/typing.html#typing.overload was that the base function doesn't get typed?

pytools/__init__.py
alexfikl: We could introduce some protocol on top of `Sequence`, but I'm not sure it's worth it?

pytools/__init__.py
inducer: I agree with that read of the language docs. At the same time, it seems that [mypy does not check the implementation unless it also has types](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading).

pytools/__init__.py
inducer: I pushed something that... somehow... mypy seems to like, that gets by without a protocol. Does that look OK to you?

alexfikl: I don't think that works though. I'm getting
```
>>> list(generate_permutations("abc"))
Traceback (most recent call last):
...
	yield tp(*perm_[:i], *original[0:1], *perm_[i:])
	~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: decoding str is not supported
```

pytools/__init__.py
alexfikl: Hm, that's not very nice at all :)) Thanks for the heads up.

pytools/__init__.py
inducer: Yikes, perfect. Untested code. :grimacing: Added a test.

inducer: And: did the protocol thing.

alexfikl: That looks good to me! Thanks for adding tests :D

PR #285, opened 2025-01-16
fix numpy scalar hash for bigendian

### LINKED ISSUE #262: `pytools/test/test_persistent_dict.py::test_{attrs,dataclass}_hashing` failures on big-endian
I'm sorry that I've missed this previously but looks like cee5027972c5c5063f64f7f6624e20d1c9069497 introduced two test regressions on big-endian:

```pytb
$ pytest pytools/test/test_persistent_dict.py::test_{attrs,dataclass}_hashing
========================================================= test session starts =========================================================
platform linux -- Python 3.12.3, pytest-8.3.2, pluggy-1.5.0
rootdir: /home/mgorny/pytools
configfile: pyproject.toml
plugins: xdist-3.6.1
collected 2 items                                                                                                                     

pytools/test/test_persistent_dict.py FF                                                                                         [100%]

============================================================== FAILURES ===============================================================
_________________________________________________________ test_attrs_hashing __________________________________________________________

    def test_attrs_hashing() -> None:
        attrs = pytest.importorskip("attrs")
    
        keyb = KeyBuilder()
    
        @attrs.define
        class MyAttrs:
            name: str
            value: int
    
>       assert (keyb(MyAttrs("hi", 1)) == "5b6c5da60eb2bd0f")  # type: ignore[call-arg]
E       AssertionError: assert '626f565387d8b006' == '5b6c5da60eb2bd0f'
E         
E         - 5b6c5da60eb2bd0f
E         + 626f565387d8b006

pytools/test/test_persistent_dict.py:593: AssertionError
_______________________________________________________ test_dataclass_hashing ________________________________________________________

    def test_dataclass_hashing() -> None:
        keyb = KeyBuilder()
    
        @dataclass
        class MyDC:
            name: str
            value: int
    
>       assert keyb(MyDC("hi", 1)) == "d1a1079f1c10aa4f"
E       AssertionError: assert 'e36c7d4306a3670a' == 'd1a1079f1c10aa4f'
E         
E         - d1a1079f1c10aa4f
E         + e36c7d4306a3670a

pytools/test/test_persistent_dict.py:569: AssertionError
======================================================= short test summary info =======================================================
FAILED pytools/test/test_persistent_dict.py::test_attrs_hashing - AssertionError: assert '626f565387d8b006' == '5b6c5da60eb2bd0f'
FAILED pytools/test/test_persistent_dict.py::test_dataclass_hashing - AssertionError: assert 'e36c7d4306a3670a' == 'd1a1079f1c10aa4f'
========================================================== 2 failed in 0.35s ==========================================================
```

Since we are using the system byte order now, the hashes are different on big-endian and little-endian systems. I suppose this can be fixed either by updating the test to have different expected values depending on `sys.byteorder`, or by mocking it to force little-endian:

```diff
diff --git a/pytools/test/test_persistent_dict.py b/pytools/test/test_persistent_dict.py
index b04d43a..cdddb37 100644
--- a/pytools/test/test_persistent_dict.py
+++ b/pytools/test/test_persistent_dict.py
@@ -1,6 +1,7 @@
 import shutil
 import sys
 import tempfile
+import unittest.mock
 from dataclasses import dataclass
 from enum import Enum, IntEnum
 from typing import Any, Dict, Optional
@@ -566,7 +567,8 @@ def test_dataclass_hashing() -> None:
         name: str
         value: int
 
-    assert keyb(MyDC("hi", 1)) == "d1a1079f1c10aa4f"
+    with unittest.mock.patch("sys.byteorder", "little"):
+        assert keyb(MyDC("hi", 1)) == "d1a1079f1c10aa4f"
 
     assert keyb(MyDC("hi", 1)) == keyb(MyDC("hi", 1))
     assert keyb(MyDC("hi", 1)) != keyb(MyDC("hi", 2))
@@ -590,7 +592,8 @@ def test_attrs_hashing() -> None:
         name: str
         value: int
 
-    assert (keyb(MyAttrs("hi", 1)) == "5b6c5da60eb2bd0f")  # type: ignore[call-arg]
+    with unittest.mock.patch("sys.byteorder", "little"):
+        assert (keyb(MyAttrs("hi", 1)) == "5b6c5da60eb2bd0f")  # type: ignore[call-arg]
 
     assert keyb(MyAttrs("hi", 1)) == keyb(MyAttrs("hi", 1))  # type: ignore[call-arg]
     assert keyb(MyAttrs("hi", 1)) != keyb(MyAttrs("hi", 2))  # type: ignore[call-arg]
```

Please let me know if I should make a PR doing either of the two.

matthiasdiener:
Fixes #262. Followup of #259, #260, #261.

Reverts #260.
Closes #261.

TODO:
- [x] add a test with mock

_Please squash_

### DIFF
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -68,7 +68,7 @@ else:
     _HAS_ATTRS = True
 
 
-IS_BIGENDIAN = sys.byteorder == "big"
+_IS_BIGENDIAN = sys.byteorder == "big"
 
 
 logger = logging.getLogger(__name__)
@@ -336,7 +336,7 @@ class KeyBuilder:
         elif hasattr(np, "float128") and key.dtype == np.dtype("float128"):
             key_hash.update(repr(float(key)).encode("utf8"))
         else:
-            if IS_BIGENDIAN:
+            if _IS_BIGENDIAN:
                 key_hash.update(np.array(key).byteswap().tobytes())
             else:
                 key_hash.update(np.array(key).tobytes())
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -219,6 +219,8 @@ class KeyBuilder:
                         # Non-numpy scalars are handled above in the try block.
                         method = self.update_for_numpy_scalar
 
+                    # numpy arrays are mutable so they are not handled in KeyBuilder
+
                 if method is None:
                     if issubclass(tp, Enum):
                         method = self.update_for_enum
@@ -324,6 +326,8 @@ class KeyBuilder:
 
     def update_for_numpy_scalar(self, key_hash: Hash, key: Any) -> None:
         import numpy as np
+        if sys.byteorder == "big":
+            key = key.byteswap()
         if hasattr(np, "complex256") and key.dtype == np.dtype("complex256"):
             key_hash.update(repr(complex(key)).encode("utf8"))
         elif hasattr(np, "float128") and key.dtype == np.dtype("float128"):
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -68,6 +68,7 @@ else:
     _HAS_ATTRS = True
 
 
+# Used to speed up the comparison in update_for_numpy_scalar and for mock testing
 _IS_BIGENDIAN = sys.byteorder == "big"
 
 
@@ -331,6 +332,9 @@ class KeyBuilder:
 
     def update_for_numpy_scalar(self, key_hash: Hash, key: Any) -> None:
         import numpy as np
+
+        # tobytes() of np.complex256 and np.float128 are non-deterministic,
+        # so we need to use their repr() instead.
         if hasattr(np, "complex256") and key.dtype == np.dtype("complex256"):
             key_hash.update(repr(complex(key)).encode("utf8"))
         elif hasattr(np, "float128") and key.dtype == np.dtype("float128"):
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -68,6 +68,9 @@ else:
     _HAS_ATTRS = True
 
 
+IS_BIGENDIAN = sys.byteorder == "big"
+
+
 logger = logging.getLogger(__name__)
 
 # NOTE: not always available so they get hardcoded here
@@ -271,6 +274,8 @@ class KeyBuilder:
 
     def update_for_int(self, key_hash: Hash, key: int) -> None:
         sz = 8
+        # Note: this must match the hash for numpy integers, since
+        # np.int64(1) == 1
         while True:
             try:
                 key_hash.update(key.to_bytes(sz, byteorder="little", signed=True))
@@ -326,7 +331,7 @@ class KeyBuilder:
 
     def update_for_numpy_scalar(self, key_hash: Hash, key: Any) -> None:
         import numpy as np
-        if sys.byteorder == "big":
+        if IS_BIGENDIAN:
             key = key.byteswap()
         if hasattr(np, "complex256") and key.dtype == np.dtype("complex256"):
             key_hash.update(repr(complex(key)).encode("utf8"))
--- a/pytools/persistent_dict.py
+++ b/pytools/persistent_dict.py
@@ -163,12 +163,6 @@ class KeyBuilder:
         may stop working as early as 2022.
 
         .. versionadded:: 2021.2
-
-    .. note::
-
-        Some key-building uses system byte order, so the built keys may not match
-        across different systems. It would be desirable to fix this, but this is
-        not yet done.
     """
 
     # this exists so that we can (conceivably) switch algorithms at some point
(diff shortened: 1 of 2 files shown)


matthiasdiener:
This is ready for a first look @inducer @mgorny

pytools/test/test_persistent_dict.py
```diff
@@ -479,6 +479,19 @@ def test_scalar_hashing() -> None:
...
+
+    import unittest.mock
+    with unittest.mock.patch("pytools.persistent_dict.IS_BIGENDIAN", True):
+        # Very crude big-endian test. Apparently, there is no way to
+        # create a numpy scalar (not an np.array!) of a specific byte order.
+        if sys.byteorder == "big":
+            # no need to swap
+            be_val = np.int64(1)
+        else:
+            be_val = np.int64(1).byteswap()
```
mgorny, line 492: This part is kinda redundant on big-endian systems, so I'm wondering: how about testing little endian there instead? i.e. something like:

```suggestion
    with unittest.mock.patch("pytools.persistent_dict.IS_BIGENDIAN", sys.byteorder == "little"):
        # Very crude test that the other endianness works as well.
        be_val = np.int64(1).byteswap()
```

mgorny (commented):
Tests pass for me on PPC64.

pytools/test/test_persistent_dict.py
matthiasdiener: Thanks, what do you think of [4ce5d88](https://github.com/inducer/pytools/pull/285/commits/4ce5d88b649b3663b303255047ff9fe408027c46)?

Out of curiosity, what is the PPC64 hardware that you are using?

mgorny: Yep, that matches my idea precisely and passes on our machine.

To be honest, I don't know exactly, I'm only an user there — apparently it's some kind of virtualized POWER9 server.

matthiasdiener: > Yep, that matches my idea precisely and passes on our machine.
> 
> To be honest, I don't know exactly, I'm only an user there — apparently it's some kind of virtualized POWER9 server.

Ah interesting, I was under the impression that many Power9s were running in LE mode.

pytools/persistent_dict.py
```diff
@@ -68,6 +68,9 @@ class RecommendedHashNotFoundWarning(UserWarning):
     _HAS_ATTRS = True
 
 
+IS_BIGENDIAN = sys.byteorder == "big"
```
inducer, line 71:
```suggestion
_IS_BIGENDIAN = sys.byteorder == "big"
```

inducer, line 71: Add a comment that this is here so it can be changed by mock testing.

pytools/test/test_persistent_dict.py
mgorny: Yeah, PPC64 is bi-endian. Since Gentoo still supports big-endian PPC64, we literally have two testing machines: one set to run LE, the other BE. In fact, it's quite a nice setup, because we can test packages on two very similar systems with different endianness, and so easily distinguish endianness-related issues from other kinds of problems that occur on BE hardware such as SPARC.

inducer (commented):
Two minor nits, otherwise good to go.

pytools/persistent_dict.py
matthiasdiener: Done in [49820c7](https://github.com/inducer/pytools/pull/285/commits/49820c79a16672e3aeecea1a715d08fa60978fe1)

pytools/persistent_dict.py
```diff
@@ -68,6 +68,10 @@ class RecommendedHashNotFoundWarning(UserWarning):
     _HAS_ATTRS = True
 
 
+# Used to speed up the comparison in update_for_numpy_scalar and for mock testing
```
inducer, line 71: Is the "speed up" you mention here in relation to redoing the comparison `sys.byteorder == "big"`? If so, I think that's sort of a questionable thing to claim.

matthiasdiener: Well, it seems to be significantly faster with the bool comparison. In fact, the speed up is the main reason - we could just patch `sys.byteorder` directly with mock.

On M1 / Python 3.11.9:

```python
import sys
from time import monotonic

bl = True

# Warmup
for i in range(10000000):
    bl == True
    sys.byteorder == "little"

start = monotonic()
for i in range(10000000):
    bl == True
print(monotonic() - start)

start = monotonic()
for i in range(10000000):
    sys.byteorder == "little"

print(monotonic() - start)
```


```console
$ python t.py
0.3181627501035109
0.41722170799039304
```

inducer: Less dramatic on x86_64:
```
0.27424564998364076
0.3040320659929421
```
Moved into a function, things get faster:
```
0.16420180699788034
0.1805221170070581
```

inducer: At any rate, LGTM!

mgorny:
Thanks a lot, to you both!

PR #293, opened 2025-03-27
Don't duplicate `Taggable` when adding already-existing tag or removing non-existing tag

majosm:


inducer:
Thanks!

pytools/tag.py
inducer, line 1: @majosm What do you think of this version? Set equality is kind of expensive to check, $O(n+m)$. Subset and intersection are cheaper, $O(n)$, especially if $n$ is small.

majosm: Sure, makes sense!

inducer:
https://pypi.org/project/pytools/2025.1.2/

PR #296, opened 2025-04-03
add a small test for hashability of bare classes

matthiasdiener:


pytools/test/test_persistent_dict.py
```diff
@@ -528,6 +528,16 @@ def update_persistent_hash(self, key_hash, key_builder):
 def test_class_hashing() -> None:
     keyb = KeyBuilder()
 
+    class WithoutUpdateMethod:
+        pass
+
+    assert keyb(WithoutUpdateMethod) == keyb(WithoutUpdateMethod)
```
inducer, line 534: Arguably, this should not work. This is a different class every time the function gets called. It might be useful to see if the class type has the same `id()` as the one available at the advertised name, before using that name for hashing.

matthiasdiener: TIL, thanks! What do you think of e35c58a? I'm worried that this may cause extra cache misses since some tags that we use might be local classes?

pytools/persistent_dict.py
```diff
@@ -244,8 +244,22 @@ def __call__(self, key: Any) -> str:
...
+            for attr in key.__qualname__.split("."):
+                resolved = getattr(resolved, attr)  # pyright: ignore[reportAny]
+            if resolved is key:
+                # Globally accessible: hash based on name
+                self.rec(key_hash,
+                    f"{key.__module__}.{key.__qualname__}".encode()
+                )
+            else:
+                # Name collision or local class; fall back
+                self.rec(key_hash, f"{key.__module__}.{key.__qualname__}_{id(key)}")
```
inducer, line 259: I think local classes should be unhashable.

PR #297, opened 2025-05-05
Go back to requiring siphash24

inducer:
Closes https://github.com/inducer/pyopencl/issues/827


inducer:
@connorjward This having been merged notwithstanding: are there any concerns about this from your end? I'll hold off on rolling a release until I've heard from you.

connorjward:
> @connorjward This having been merged notwithstanding: are there any concerns about this from your end? I'll hold off on rolling a release until I've heard from you.

We were already using the siphash optional dependency to avoid the warning. I think this will be fine.

PR #300, opened 2025-05-09
Maint 2025 05

inducer:


pyproject.toml
```diff
@@ -112,6 +109,10 @@ reportUnnecessaryIsInstance = "none"
 reportUnusedCallResult = "none"
 reportExplicitAny = "none"
 
+# This reports even cycles that are qualified by 'if TYPE_CHECKING'. Not what
+# we care about at this moment.
+# https://github.com/microsoft/pyright/issues/746
+reportImportCycles = "none"
 
 [tool.mypy]
 python_version = "3.10"
```
matthiasdiener, line 118: Not suer if it matters, but the mypy CI runs the mypy test with 3.13.

inducer: Thanks! I've added analogous configuration for basedpyright.

PR #321, opened 2025-08-31
Small fixes for pytential

alexfikl:


pytools/obj_array.py
```diff
@@ -273,15 +273,21 @@ def __matmul__(
...
+                /) -> ObjectArray2D[T]: ...
         @overload
         def __matmul__(
                     self: ObjectArrayND[T],
                     other: ObjectArrayND[T],
                 /) -> ObjectArrayND[T] | T: ...
 
         @property
-        def flat(self) -> ObjectArray1D[T]: ...
+        def flat(self) -> Iterator[T]: ...
```
alexfikl, line 288: `ndarray.flat` actually returns a `np.flatiter`, but that's generic on something with a bound on `NDArray[Any]`, so we can't really use it here. This should still be good enough?

pytools/obj_array.py
```diff
@@ -439,7 +445,7 @@ def sum(
 
 def sum(
             array: ObjectArrayND[T],
-            axis: int | None,
+            axis: int | None = None,
```
alexfikl, line 448: Did this not have a default on purpose? I would expect it to match `np.sum`.

alexfikl:
> Thx! LMK if you think this should be a release right away.

No, there's no hurry. It just fixes a few more errors in there.

PR #323, opened 2025-09-02
Fix infinite recursion in `obj_array_vectorize_n_args`

majosm:


majosm:
@inducer Should I just update the bpr baseline to fix CI? Doesn't look like `vectorize_n_args` has types yet.

inducer:
https://pypi.org/project/pytools/2025.2.4/

PR #356, opened 2026-06-14
Fix new pyright errors

alexfikl:
The new `optype==0.18.0` requires Python 3.12 for typing (mostly seems to use `typing.TypeAliasType` now). This bumps the version in `pyproject.toml` for `basedpyright` and fixes some small errors.

From what I could tell, a bunch of other projects using `optype` also have the same issues :\

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
                 python-version: '3.x'
         -   name: "Main Script"
             run: |
-                EXTRA_INSTALL="basedpyright numpy attrs orderedsets pytest mpi4py matplotlib optype"
+                EXTRA_INSTALL="basedpyright numpy attrs orderedsets pytest mpi4py matplotlib optype<0.18"
 
                 sudo apt update
                 sudo apt -y install libopenmpi-dev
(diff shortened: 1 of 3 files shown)


inducer:
Don't know what the right path is here. I'm not sure I'm ready to drop 3.10. It's [not EOL](https://devguide.python.org/versions/), and Firedrake allows 3.10+ as well:

https://github.com/firedrakeproject/firedrake/blob/6386785d7dd168fc0c1392a6558247b8496de246/pyproject.toml#L15

And I appreciate that basedpyright warns me if I try to use language features that aren't in that. So I think this might be a case of upper-bounding optype instead? What do you think?

alexfikl:
> And I appreciate that basedpyright warns me if I try to use language features that aren't in that. So I think this might be a case of upper-bounding optype instead? What do you think?

I agree that it's nice to have some warnings about language features and things. We can try? The main worry is that may be a pain going forward if `optype<0.18` / `numpy` / `scipy` start diverging somehow, but we'll see..

inducer:
Thanks! Yes, let's try that and see how it goes.

inducer:
Apparently there's a grand plan behind this madness:

https://scientific-python.org/specs/spec-0000/

PR #358, opened 2026-06-22
Update baseline

alexfikl:
[New numpy is out](https://github.com/numpy/numpy/releases/tag/v2.5.0) with typing changes :grin: All of the new errors seem to complain that some annotation is partially unknown..

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -3836,30 +3836,6 @@
                 }
             },
             {
-                "code": "reportUnknownMemberType",
-                "range": {
-                    "startColumn": 11,
-                    "endColumn": 29,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportUnknownMemberType",
-                "range": {
-                    "startColumn": 11,
-                    "endColumn": 34,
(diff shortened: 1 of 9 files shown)


alexfikl:
Hm, most if not all of these errors seem to be due to `numpy` 2.5 using typing stuff specific for Python 3.12, e.g.
```python
@runtime_checkable
class _SupportsArray[DTypeT: np.dtype](Protocol):
    def __array__(self) -> np.ndarray[Any, DTypeT]: ...
```
This makes the whole protocol `Unknown` for pyright with a bound on Python 3.10.

inducer:
Ugh. I think the 0.18 bound for optype was the wrong call then. Sorry! Let's unwind that.

inducer:
(And, consequently, bump the basedpyright python version to 3.12.)

alexfikl:
> Ugh. I think the 0.18 bound for optype was the wrong call then. Sorry! Let's unwind that.

Yeah.. I'm guessing they were in cahoots about that and made the releases match :\

inducer:
Thank you!

PR #363, opened 2026-08-28
Fix new ruff errors

alexfikl:
The new version started being more verbose about `unnecessary-dunder-call` (which we use quite a bit for frozen dataclasses) and `pytest-parameter-with-default-argument` (which we use for `visualize=True` and things like that), so it will be a bit of a pain :facepalm:


inducer:
Thx!

PR #365, opened 2026-09-16
Bump platformdirs version

alexfikl:
`platformdirs` relented and added support for `XDG_*` variables in https://github.com/tox-dev/platformdirs/releases/tag/4.6.0. 

That version supports Python 3.10+, so we should be good to force it.


inducer:
Thx!

PR #367, opened 2026-09-19
Add a timeout to download_from_web_if_not_present

alexfikl:
Wanted to add a timeout to `download_from_web_if_not_present` for something.. ended up also reworking `invoke_editor` a bit: clean up the temporary directory and handle `EDITOR` set to something like `emacs -nw`.


inducer:
Thx!

